PR #5, opened 2018-10-11
VBR Crash Fix

dseah-intel:
- Fixed crash due to hlRateControlHistorgramQueue overflow
- Fixed UpdateHistogramQueueEntry() function locking on the wrong mutex


tianjunwork:
@dseah-intel Thx for a quick fix! I tested, it doesn't crash for 512x512. So it should be better that commit msg be more specific on what failed.

dseah-intel:
@hassount Hi Hassene, the latest push reflects the changes for error checking

PR #7, opened 2018-10-19
Build: change the file mode for linux build.

mypopydev:
Change the linux build file mode and update READEME.

Signed-off-by: Jun Zhao <jun.zhao@intel.com>


mypopydev:
@tianjunwork After this commit with mode change, i don't think need to the command "set fileformat=unix" for build.sh

PR #15, opened 2018-10-24
Terminate sample app with unrecognized parameters

tianjunwork:
Check spelling of parameters start with "-".
Terminate execution with unrecognized parameters.

### DIFF
--- a/Source/App/EbAppConfig.c
+++ b/Source/App/EbAppConfig.c
@@ -1353,6 +1353,7 @@ EB_ERRORTYPE ReadCommandLine(
             printf(" %s ", cmd_copy[cmd_copy_index]);
         }
         printf("\n\n");
+        return_error = EB_ErrorBadParameter;
     }
 
     if (return_error == EB_ErrorNone){


hassount:
@tianjunwork 
So this change will cause the encoder not to start encoding if a wrong parameter has been detected with the existing checks that the application has ? 
If such is the case, then we should change the name of the pull request / commit and make it reflect the functionality added. 

Or add a newer functionality allowing the application to spell check for other characters. At this point the following command lines will survive through the encode:
-i d:\stream\a.yuv -w 864 -h 480 bit-depth 8
-i d:\stream\a.yuv -w 864 -h 480 ~bit-depth 8

tianjunwork:
@hassount I changed the name of the pull request / commit. 
Since the sample code accept multiple value for one token(e.g. multiple config file for multi-channel), it only checks spelling of tokens start with "-", not yet everything between two tokens. "~bit-depth 8" is considered as values for -h.
Users usually already noticed that all the tokens starts from "-", but may have typo like "--bit-depth" or "-bit_depth" or "-bit-dpeth". The new logic helps them find the error earlier before getting frustrated with the wrong bitstream. It is good enough for now.

hassount:
@tianjunwork 
Great, thank you Jun.

hassount:
@lijing0010 
If you're good with these changes, we can merge them.

PR #16, opened 2018-10-25
lib/system/EbEncDecProcess: Coding style clean up.

mypopydev:
Signed-off-by: Jun Zhao <jun.zhao@intel.com>

### DIFF
--- a/Source/Lib/System/EbEncDecProcess.c
+++ b/Source/Lib/System/EbEncDecProcess.c
@@ -371,7 +371,7 @@ static EB_ERRORTYPE ApplySaoOffsetsLcu(
             }
         }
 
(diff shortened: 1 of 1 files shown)


mypopydev:
@tianjunwork Do you want to give a big commit for coding style? I don't think is a good idea, because nobody like a  supersize commit, the other thing is, if the submitter missing something in the supersize commit, the reviewer is hard to review and found the issue in the commit. So I think we can formatting current source code based on file  granularity if we found the same format issue.

juntianwork:
There are many tools to format source code automatically in batch. There will be too many commits if you want to do on file granularity.

mypopydev:
I know have the tools for auto format, but I don't think anybody like a supersize commit, this is my point.

PR #20, opened 2018-11-05
Fix support for non multiple of 8 resolutions

hassount:
Applicable for both 10bit and 8bit

### DIFF
--- a/Source/Lib/System/EbEncHandle.c
+++ b/Source/Lib/System/EbEncHandle.c
@@ -1818,10 +1818,10 @@ void CopyApiFromApp(
     sequenceControlSetPtr->chromaHeight = sequenceControlSetPtr->maxInputLumaHeight >> 1;
 
     // Configure the padding
-    sequenceControlSetPtr->leftPadding = MAX_LCU_SIZE + 4;
-    sequenceControlSetPtr->topPadding = MAX_LCU_SIZE + 4;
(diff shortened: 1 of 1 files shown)


Source/Lib/System/EbEncHandle.c
```diff
@@ -2172,11 +2172,19 @@ static EB_ERRORTYPE VerifySettings(\
         printf( "Error Instance %u: Base Layer Switch Mode 1 only when Prediction Structure is Random Access\n", channelNumber + 1);
         return_error = EB_ErrorBadParameter;
     }
-
-	if (sequenceControlSetPtr->maxInputLumaWidth % 8) {
-        printf("Error Instance %u: Only multiple of 8 width is supported in this release \n",channelNumber+1);
+    if ((sequenceControlSetPtr->maxInputLumaWidth % 8) && config->compressedTenBitFormat == 1) {
```
tianjunwork: Please clean up similar logic several lines below:
if (sequenceControlSetPtr->maxInputLumaWidth % 8 != 0 && config->encoderBitDepth == 10)
 if (sequenceControlSetPtr->maxInputLumaHeight % 8 != 0 && config->encoderBitDepth == 10) {

hassount: The compressed 10-bit format can not be a non-multiple of 8 for now.

hassount: compressedTenBitFormat and encoderBitDepth are not the same and compressedTenBitFormat requires a multiple of 8 by design

PR #21, opened 2018-11-05
Fix build failure for O0 optimization level

hassount:
Added static for functions used within the same file


tianjunwork:
@hassount you mean O0 Optimization level?
Other than the title, I am OK with the patch.

PR #36, opened 2018-12-12
API Cleanup for Gstreamer plugin

hassount:


Source/App/EbAppMain.c
```diff
@@ -323,7 +305,7 @@ int main(int argc, char* argv[])
         }
         else {
             printf("Error in configuration, could not begin encoding! ... \n");
-            printf("Run %s -help for a list of options\n", argv[0]);
```
tianjunwork: The old one is better to avoid the hardcoded name.

hassount: Addressed in last commit

hassount:
Merge conflict fixed and merged in #39

PR #37, opened 2018-12-13
External API

fhvwy:
Prompted by the submission to FFmpeg, this cleans up the external API to be usable with other packages (mostly this is fixing namespace problems).

* The patches removing unused things could instead add suitable namespacing if there are external dependencies wanting them.
* In patch 6 (making the internal build configuration internal), the location chosen to put the new header is quite arbitrary - it just needed to go somewhere separate from both the library and the app because it's used by both.  That can be moved or renamed if desired.
* Patch 11 (adding version script) seemed like the easiest way to hide the random symbols which are currently injected into the global namespace when dynamic linking.  It could probably also be fixed by being more consistent with the visibility options on every symbol (as is partially present in the existing code), but this method seemed simpler and also does not require continuing maintenance as new symbols are added.
* The change to SONAME in patch 12 does breaks existing dynamic links, but this is correct given that the API is changed by the preceding fixes.
* Only tested on Linux with FFmpeg.

Some other thoughts on API:
* It might be clearer to use fixed-size types (intNN_t, uintNN_t) throughout the API.  While the current intended targets (x86 32- and 64-bit) do match for all the types used, that would make it explicit.
* It's not clear what the use of the FOO_SLICE variables (renamed in patch 9) is - they don't match the slice types defined in the standard, nor do they tell you the frame type.  It looks like the main intended use is to identify IRAPs, but this isn't obvious - IDR is easy, but what about CRA as opposed to other intra-only slices?  If IRAPs are the only use for them, then maybe a single buffer flag would be a simpler method.
* While fixed-size user-allocated buffers for output made sense in OpenMAX IL for hardware codecs, it doesn't really for this software codec.  They should probably be allocated by the library somehow to avoid any problems around how big they need to be.
* I don't think asking for SCHED_FIFO priority unconditionally is a good idea - it's not intended for processes that run all the time.  It might make sense in some realtime use-cases where the CPU will be idle between frames, but when doing offline transcodes it immediately hits the realtime scheduling limit such that it probably ends up getting less CPU time than a SCHED_OTHER process would.  Maybe an option would be better for cases where this is actually wanted?  The prompt for the user to run as root is also inappropriate - this can be set more safely for normal users via /etc/security/limits.conf or polkit.
* The EB_ASM types seem a bit confused - EB_ASM_TYPE_TOTAL is used internally to size some arrays, but it's smaller than EB_ASM_AVX512?  I suspect EB_ASM_TYPE_TOTAL shouldn't be in the external API at all.


dseah-intel:
@fhvwy I have create a pull-request to the external-api branch updated to the master branch from my fork. Thank you.

fhvwy:
@hassount  Rebased.

> These macros (apart from DEADLOCK_DEBUG) are removed in #36 so we can remove them from the list

Yep, only `DISPLAY_MEMORY` remains - I've treated that here in the same way as `DEADLOCK_DEBUG`.

> > It's not clear what the use of the FOO_SLICE variables (renamed in patch 9) is - they don't match the slice types defined in the standard, nor do they tell you the frame type. It looks like the main intended use is to identify IRAPs, but this isn't obvious - IDR is easy, but what about CRA as opposed to other intra-only slices? If IRAPs are the only use for them, then maybe a single buffer flag would be a simpler method.

> The library supports only two types of intra: IDR or CRA so are you suggesting that we should add another slice type that won't be supported ?

Hmm.  Maybe it shouldn't be called "slice type", since it isn't really the slice type that is being indicated (CRA, IDR and non-IRAP intra access units all use I slices)?  Perhaps it could be named something like "picture type" or "access unit type" and then have values matching that: `EB_PICTURE_TYPE_IDR`,  `EB_PICTURE_TYPE_CRA`.

PR #43, opened 2018-12-18
propagate external and main CFlags settings to the libs

ph0b:
I've done these changes to keep CFlags from external and main project settings and only append to them in the libs as needed, and fix MinGW compilation.
I've validated these with MinGW GCC 7.3, 8.2, and MSVC 2017 on Windows, GCC 5.4 on Linux.
that also fixes the issue of the additional Release directory on Windows, related here: https://github.com/intel/SVT-HEVC/pull/42


CMakeLists.txt
```diff
@@ -3,10 +3,12 @@ set(project_name "svt-hevc")
 
 project(${project_name} C ASM_NASM)
 
-if(NOT CMAKE_BUILD_TYPE)
-    set(CMAKE_BUILD_TYPE Release)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    if(NOT CMAKE_BUILD_TYPE)
+        set(CMAKE_BUILD_TYPE Release)
+    endif()
+    message(STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE}")
```
tianjunwork: I will revert my fix for the binary folder

PR #45, opened 2018-12-20
CMakeLists.txt improvements

xwu2git:
(1)	Installed include files under include/svt-hevc to avoid polluting the include directories. This is needed if you want to install to system path: /usr/include.
(2)	Reduced cmake minimum version to v2.8 (to make CentOS happy.)
(3)	Added STATIC to libraries so that –DBUILD_SHARED_LIBS=ON does not cause trouble.
(4)	Fixed binaries and pkg-config files installing to the proper system path (specified by CMAKE_INSTALL_LIBDIR.)

Tested under CentOS 7.5,  Ubuntu 16.04 and Ubuntu 18.04.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 2.8)
 set(project_name "svt-hevc")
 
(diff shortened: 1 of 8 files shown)


xwu2git:
Updated to the latest master.

PR #72, opened 2019-01-11
Limit maximum lookAheadDistance

dseah-intel:
- Fix adding extra lookAheadDistance for buffers
- Limit lookAheadDistance


dseah-intel:
@tianjunwork Hi Jun, I have fixed the config file. Thank you.

PR #75, opened 2019-01-12
#define  TILES    1  //enable TILES.

chkngit:
hevc tiles

### DIFF
--- a/Source/API/EbApi.h
+++ b/Source/API/EbApi.h
@@ -324,6 +324,10 @@ typedef struct EB_H265_ENC_CONFIGURATION
      * Default is null.*/
     uint8_t                 useQpFile;
 
+#if 1//TILES
+    uint8_t                 tileColumnCount;
+    uint8_t                 tileRowCount;
+#endif
 
     // Deblock Filter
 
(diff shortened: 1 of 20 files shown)


Source/API/EbApi.h
```diff
@@ -151,7 +151,10 @@ extern "C" {
         // Quantization
         uint32_t                qp;
         uint8_t                 useQpFile;
-
+#if 1//TILES
```
tianjunwork: you meant "#if TILES"?

chkngit: yes, #define TILES , is defined in definitions.h , and it is not visible in the app, so I left it ON, but commented out to be able to track all changes when searching for TILES

tianjunwork:
@chkngit Please fix the conflict and keep the original coding style.

Source/API/EbApi.h
tianjunwork: "#if 1//TILES" is not necessary and confusing. "#if TILES" not necessary either. A internal parameter can be defined to enable/disable tile based on TileRowCount/TileColumnCount value.
e.g. if both TileRowCount/TileColumnCount > 1, tile is enabled.

PR #77, opened 2019-01-15
Add 422/444 in header file

lijing0010:

### DIFF
--- a/Source/API/EbApi.h
+++ b/Source/API/EbApi.h
@@ -92,6 +92,13 @@ extern "C" {
 
 #define EB_BUFFERFLAG_EOS 0x00000001
 
+typedef enum EB_COLOR_FORMAT {
+    EB_YUV400,
+    EB_YUV420,
+    EB_YUV422,
+    EB_YUV444
+} EB_COLOR_FORMAT;
+
 // For 8-bit and 10-bit packed inputs, the luma, cb, and cr fields should be used
 //   for the three input picture planes.  However, for 10-bit unpacked planes the
 //   lumaExt, cbExt, and crExt fields should be used hold the extra 2-bits of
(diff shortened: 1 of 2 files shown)


Source/API/EbApi.h
```diff
@@ -92,6 +92,13 @@ extern "C" {
 
 #define EB_BUFFERFLAG_EOS 0x00000001
 
+typedef enum EB_COLOR_FORMAT {
+    EB_YUV400,
+    EB_YUV420,
+    EB_YUV422,
+    EB_YUV444
+} EB_COLOR_FORMAT;
```
tianjunwork: better name with EB_CHROMAFORMAT_YUVXXX? 
typedef enum EB_COLOR_FORMAT  {
   EB_CHROMAFORMAT_YUV400 = 0,
  EB_CHROMAFORMAT_YUV_420        = 1,
  EB_CHROMAFORMAT_YUV_422        = 2,
  EB_CHROMAFORMAT_YUV_444        = 3,
  } EB_COLOR_FORMAT;

Source/API/EbApi.h
```diff
@@ -138,6 +145,7 @@ typedef struct EB_H265_ENC_CONFIGURATION
     int32_t                 frameRateNumerator;
     int32_t                 frameRateDenominator;
     uint32_t                encoderBitDepth;
+    EB_COLOR_FORMAT         colorFormat;
```
tianjunwork: where is it used?

Source/API/EbApi.h
lijing0010: it's not only about chroma, '4' denotes for luma actually

Source/API/EbApi.h
lijing0010: I'm enabling 422/444 support, so it will be used for following patches

tianjunwork: where are the following patches? Since it's related, the PR can still be used to add the feature.

PR #79, opened 2019-01-22
Fix bitstream mismatch for un-ref B/P frames

lijing0010:
Even if we enable DLF at application level, no DLF will be done for B/P
frames which are not used as a reference. However in previous
implementation slice_deblocking_filter_disabled_flag is not set
correctly for such case


lijing0010:
No need for this.

PR #90, opened 2019-02-01
FFmpeg Plugin: Fix the invalid parameter causing segmentation fault issue

jsunintel:


jsunintel:
Fixed https://github.com/intel/SVT-HEVC/issues/87.

PR #91, opened 2019-02-01
Fix the multi-thread issue while dumping recon

lijing0010:
In previous implementation, some recon frame may get lost. The buf with
EOS may not be the last one due to multi-threading


Source/Lib/Codec/EbEncDecProcess.c
```diff
@@ -1730,9 +1718,21 @@ static void ReconOutput(
...
+    //   the encoder might not properly terminate.
+    EbBlockOnMutex(encodeContextPtr->terminatingConditionsMutex);
+
+    // START READ/WRITE PROTECTED SECTION
+    if (encodeContextPtr->totalNumberOfReconFrames == encodeContextPtr->terminatingPictureNumber) {
+        outputReconPtr->nFlags = EB_BUFFERFLAG_EOS;
+    }
+    encodeContextPtr->totalNumberOfReconFrames++;
+
     // Post the Recon object
```
tianjunwork: Are EbGetEmptyObject and  EbPostFullObject need to be protected by terminatingConditionsMutex?
If not, only protect totalNumberOfReconFrames is enough.

PR #97, opened 2019-02-07
Fix kw issues

dseah-intel:
- Removed self-assigning code
- Replaced fopen to open

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -629,32 +629,34 @@ EB_ERRORTYPE EbSetThreadManagementParameters(
 
     int fd = open("/proc/cpuinfo", O_RDONLY | O_NOFOLLOW, "rt");
     struct stat file_stat;
-    if (fd >= 0 && fstat(fd, &file_stat) != -1 && S_ISREG(file_stat.st_mode) != 0) {
-        int processor_id = 0, socket_id = 0;
-        char line[128];
-        int bytes = 1;
-        while (bytes > 0) {
-            bytes = read(fd,line,128);
-            if (bytes > 0) {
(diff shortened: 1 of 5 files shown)


Source/Lib/Codec/EbEncHandle.c
```diff
@@ -625,33 +627,42 @@ EB_ERRORTYPE EbSetThreadManagementParameters(
...
-                while(*p < '0' || *p > '9') p++;
-                socket_id = strtol(p, NULL, 0);
-                if (socket_id < 0 || socket_id > 15) {
-                    fclose(fin);
-                    return EB_ErrorInsufficientResources;
+    int fd = open("/proc/cpuinfo", O_RDONLY | O_NOFOLLOW, "rt");
+    if (fd > 0) {
+        struct stat file_stat;
+        if (fstat(fd, &file_stat) != -1) {
+            if (S_ISREG(file_stat.st_mode) != 0) {
```
tianjunwork: these three if() can be combined to reduce nest if()

dseah-intel: the ifs depend on the value in the functions of the previous ifs

tianjunwork: struct stat file_stat; 
if(fd >=0 && fstat(fd, &file_stat) ==0 && S_ISREG(file_stat.st_mode) != 0)
with && all the functions are executed.

dseah-intel: I did not know that. Will promptly be fixed :)

dseah-intel: Done

dseah-intel: I realized that if(fd >= 0) needs to be separate or else it would not hit close(fd) when fstat(fd, &file_stat) !=0

PR #114, opened 2019-02-23
Travis: Rewrite with Windows and properly reflect macOS build failing

1480c1:
Similar to: <https://github.com/OpenVisualCloud/SVT-AV1/pull/117>
<https://github.com/OpenVisualCloud/SVT-VP9/pull/14>

Adds a MSVC Release build and a Ninja debug build.
The ninja build is allowed to fail, meaning if the final program fails, it will not mark the whole commit as failing.

I put it there because I'm interested in using mingw and ninja in the future, but I don't want to have to do the work that travis does on my own computer every update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
         - cd Build
         - cmake $TRAVIS_BUILD_DIR -G"Ninja" -DCMAKE_BUILD_TYPE=Release
         - ninja -j0
-        - $TRAVIS_BUILD_DIR/Bin/Release/Release/SvtHevcEncApp.exe -help
+        - $TRAVIS_BUILD_DIR/Bin/Release/SvtHevcEncApp.exe -help
       after_failure: cat $TRAVIS_BUILD_DIR/Build/windows/CMakeFiles/CMakeOutput.log
     - Windows Clang:
       name: windows mingw-clang build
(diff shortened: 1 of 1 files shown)


1480c1:
If the checks complete, could I just format the yml a bit better? The indention is throwing me off real bad.

EwoutH:
I would suggest removing the MSVC build, since AppVeyor already does exactly that.

The Ninja build job is a nice addition. We could also consider moving the Ninja build to AppVeyor, since Travis is already slower due to the number of jobs (only 4 are executed simultaneous).

1480c1:
> I would suggest removing the MSVC build, since AppVeyor already does exactly that.
> 
> The Ninja build job is a nice addition. We could also consider moving the Ninja build to AppVeyor, since Travis is already slower due to the number of jobs (only 4 are executed simultaneous).

I see, I did not know that part, I only added it since we had a linux and macOS release build there. I will remove it.

1480c1:
> The Ninja build job is a nice addition. We could also consider moving the Ninja build to AppVeyor, since Travis is already slower due to the number of jobs (only 4 are executed simultaneous).

I personally would not recommend to move the ninja build to AppVeyor since I do not think that mingw gcc + ninja is an actual supported build configuration right now for any of the SVT-* and usually ends up failing. I have tried mingw clang + ninja, and that seems to work though.

1480c1:
I'm going to close these until MacOS builds are fixed. I'll reopen when they are.

PR #115, opened 2019-02-24
Use all cores if /proc is not mounted

jbeich:
The code for managing thread affinity isn't portable. While DragonFly and FreeBSD support similar API to Linux their `/proc` implementation doesn't expose CPU topology. Let's restore pre-#46 behavior.


tianjunwork:
@jbeich Thanks for your contribution.

PR #117, opened 2019-02-26
Set and check correct MainEXT profile for YUV422, YUV444 or 10Bits cases

kelvinhu325:


Source/Lib/Codec/EbEncHandle.c
```diff
@@ -2558,6 +2558,12 @@ static EB_ERRORTYPE VerifySettings(\
 		return_error = EB_ErrorBadParameter;
 	} 
 
+    if((config->encoderColorFormat >= EB_YUV422 || config->encoderBitDepth > 8) && config->profile != 4)
+    {
```
lijing0010: I remember main10(420 10bit), the profile can be 1, not 4

Source/SimpleApp/EbSimpleAppContext.c
```diff
@@ -172,6 +172,11 @@ static EB_ERRORTYPE CopyConfigurationParameters(
     callbackData->ebEncParameters.sourceHeight      = config->sourceHeight;
     callbackData->ebEncParameters.encoderBitDepth   = config->encoderBitDepth;
     callbackData->ebEncParameters.encoderColorFormat       = config->encoderColorFormat;
+    if((config->encoderColorFormat >= EB_YUV422 || config->encoderBitDepth > 8) && config->profile != 4)
```
lijing0010: ditto

Source/Lib/Codec/EbEncHandle.c
kelvinhu325: Yes, this is not correct, 420 10bit should set profile to be main10 or 2

PR #120, opened 2019-02-27
install Windows version to C:\svt-hevc by default

ph0b:
CMake's default "C:\Program Files" location breaks pkg-config and tools using it because there is a space in it.

### DIFF
--- a/Build/windows/generate_vs17.bat
+++ b/Build/windows/generate_vs17.bat
@@ -1,5 +1,5 @@
 :: Copyright(c) 2018 Intel Corporation 
 :: SPDX-License-Identifier: BSD-2-Clause-Patent
 
-cmake ../.. "-GVisual Studio 15 2017 Win64" -DCMAKE_ASM_NASM_COMPILER="yasm.exe" -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
+cmake ../.. "-GVisual Studio 15 2017 Win64" -DCMAKE_ASM_NASM_COMPILER="yasm.exe" -DCMAKE_INSTALL_PREFIX=%SYSTEMDRIVE%\svt-encoders -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
 pause


Build/windows/generate_vs17.bat
```diff
@@ -1,5 +1,5 @@
 :: Copyright(c) 2018 Intel Corporation 
 :: SPDX-License-Identifier: BSD-2-Clause-Patent
 
-cmake ../.. "-GVisual Studio 15 2017 Win64" -DCMAKE_ASM_NASM_COMPILER="yasm.exe" -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
-pause
\ No newline at end of file
+cmake ../.. "-GVisual Studio 15 2017 Win64" -DCMAKE_ASM_NASM_COMPILER="yasm.exe" -DCMAKE_INSTALL_PREFIX=%SYSTEMDRIVE%\svt-hevc -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
+pause
```
tianjunwork: With consideration that there are other svt encoders, is it better to be just "C:\svt"?

ph0b: as long as there is no name collision with the other libs, that include files stay under install_prefix/include/svt-hevc for example, yes I agree it can be just C:\svt. Some may argue C:\svt is a bit too generic so maybe C:\svt-encoders will be better, I'll leave that decision up to the maintainers.

tianjunwork: How about C:\svt-encoders-install?

1480c1: Personally, I would prefer C:\svt or C:\svt-encoders for all of them as it would be easier to remember where they are if I need to call them.

tianjunwork: How about C:\svt-install. I am just afraid people use "C:\svt or C:\svt-encoders" as their source code folder already.

1480c1: Probably better to use -encoders since I have a bad habit of cloning repos directly to root and might get them mixed up

tianjunwork: @ph0b let's go ahead with C:\svt-encoders? Thx for your contribution.

ph0b: ok!

PR #125, opened 2019-03-01
Fix compiling error with 'Intel C++ 19.0 compiler'.

luofalei:
Solving compiling errors with `Intel C++ 19.0 compiler`.


tianjunwork:
@luofalei Intel compiler was removed some time ago from CMake file. Do you mind posting the instruction here that others would like to follow?

luofalei:
@tianjunwork I generated the `Visual Studio` Projects with cmake firstly, and then I changed the `Platform Toolset` to `Intel comiler`.

By fixing these errors, the builder works well.

PR #130, opened 2019-03-04
FFmpeg Plugin: Add the VMAF optimized mode to "-tune" and the "-hdr" parameter

jsunintel:


jsunintel:
Solved https://github.com/intel/SVT-HEVC/issues/104 and https://github.com/intel/SVT-HEVC/issues/123.

Issue #123, opened 2019-03-01
Request VMAF support( -tune 2) in SVT-HEVC ffmpeg plugin


jianbingzhu:

tianjunwork:
@jsunintel Please help with this request, thanks.

Issue #104, opened 2019-02-17
HDR config for ffmpeg plugin?


pryg-skok:
Have you got any plans for it? 
Anyway can you help to configure it in current version of plugin?

jsunintel:
Which HDR?

pryg-skok:
Sorry, I thought there no place for misunderstanding. Of course I mean High Dynamic Range metainformation forwarding from encoder input to output. There is option in encoder, but I can't specify it via ffmpeg plugin.

tianjunwork:
Hi @jsunintel is this option going to be exposed to ffmpeg plugin?

jsunintel:
> 
> 
> Hi @jsunintel is this option going to be exposed to ffmpeg plugin?

Will consider it, but being quite occupied right now.

PR #133, opened 2019-03-04
FFmpeg Plugin: Add the VMAF optimized mode to "-tune" and the "-hdr" parameter

jsunintel:


jsunintel:
Solved #104 and #123.

PR #134, opened 2019-03-04
[ffmpeg_plugin] Sync pkg-config include directory

jbeich:
pkg-config is supposed to provide all the flags required to build against this library. Regressed by #111.
x265 with `-DENABLE_SVT_HEVC=ON` still builds fine because it uses [CMake module](https://github.com/videolan/x265/blob/master/source/cmake/Findsvthevc.cmake) rather than pkg-config to find a directory that contains EbApi.h.


tianjunwork:
Hi @jsunintel, please review this change. Thx.

jsunintel:
OK

jbeich:
Obsolete after #149.

Issue #124, opened 2019-03-01
4K 10bit decoding bug in EncoderMode=12


luofalei:
I tried to encode in the fastest mode for 4K 10bit sequence, like `Netflix_FoodMarket2`. There seems to have some bugs in the encoded bit-stream which made the `ffmpeg` decoder to output wrong images in the bottom rows in some frames.

The sequence could be downloaded from here:
https://media.xiph.org/video/derf/
And I used `ffmpeg` to convert it from `y4m` to `yuv420p10` format.

The decoding image and bit-stream is public here.
https://github.com/luofalei/svt-hevc-bug-info/tree/master/issue124

![decoding image sample][1]

  [1]: https://github.com/luofalei/svt-hevc-bug-info/raw/master/issue124/svt-hevc-encMode12-decoding-error.png

luofalei:
The decoding image and bit-stream is public here.
https://github.com/luofalei/svt-hevc-bug-info/tree/master/issue124

![decoding image sample][1]

  [1]: https://github.com/luofalei/svt-hevc-bug-info/raw/master/issue124/svt-hevc-encMode12-decoding-error.png

luofalei:
I tested the older versions and found it works good in 
```
SHA-1: 39574132c9b0cd83e5589fbb8da49d3a20e37973, Branch RevNo: 81

* Fix the frame rate issue (#108)

Signed-off-by: Jing Li <jing.b.li@intel.com>
```
However, the above decoding error appears in the later version such as
```
SHA-1: 51c5f427ad28320ec6765f9406ae02b25ca4c93d, Branch RevNo: 84

* Add 422/444 support (#105)

* Add 422/444 support in buf desc and PCS

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable 422/444 in ResourceCoordination Kernel

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable 422->420 chroma down sampling in MT kernel

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable denoise AVX optimization for 422/444

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable 422/444 in ModeDecision Kernel

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable 422/444 8/10bit support

Signed-off-by: Jing Li <jing.b.li@intel.com>

* Enable 422/444 support at simple app

Signed-off-by: Jing Li <jing.b.li@intel.com>

[...]

PR #141, opened 2019-03-11
Fix YUV420p10 format encoding error.

luofalei:
#Issue 124

This is a bugfix for #124 . Bugs in entropy coding was fixed.


tianjunwork:
Hi @lijing0010, could you review also? And check if it works well with yuv444/422?

lijing0010:
Thanks Falei, just to double confirm with you.
The root cause is the uninitialized value, not the entropy itself, right?

I saw in your local repo, you also changed something in EbEntropyCoding.c in function EncodeQuantizedCoefficients_xxx(), which seems not right to me.

luofalei:
The root cause is the uninitialzed value `cbCbf2` and `crCbf2` leading to error in entropy coding. 

The new force-pushed commit (https://github.com/intel/SVT-HEVC/commit/000933fc8d2ba4ecbb9872f7953cd24fd7a9d4c8) can solve this bug.

PR #146, opened 2019-03-13
Change nalu file setting

hwrenx:
Maybe svt-hevc could only enable useNaluFile  flag when input nalu file exists.  With throwing a warning instead of an error may be better that we won't stop encoding if we just missing a nalu file : )

### DIFF
--- a/Source/App/EbAppConfig.c
+++ b/Source/App/EbAppConfig.c
@@ -152,7 +152,10 @@ static void SetNaluFile(const char *value, EbConfig_t *cfg)
 {
     if (cfg->naluFile) { fclose(cfg->naluFile); }
     FOPEN(cfg->naluFile, value, "rb");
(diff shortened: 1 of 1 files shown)


Source/App/EbAppConfig.c
```diff
@@ -152,7 +152,10 @@ static void SetNaluFile(const char *value, EbConfig_t *cfg)
 {
     if (cfg->naluFile) { fclose(cfg->naluFile); }
     FOPEN(cfg->naluFile, value, "rb");
-    cfg->useNaluFile = EB_TRUE;
+    if (cfg->naluFile)
+        cfg->useNaluFile = EB_TRUE;
+    else 
+        printf("Nalu file: %s does not exist, won't use.", value);
```
tianjunwork: Thanks for your contribution.
Please add "\n" at the end.
I think it is better to add "Error: " at the beginning.

PR #149, opened 2019-03-14
FFmpeg-Plugin: Revert the header file names' change

jsunintel:
SVT AV1 has changed the duplicated header file names, so I am reverting the change in SVT HEVC.


jsunintel:
@jbeich

PR #170, opened 2019-04-01
EbEncHandle: Use __VERSION__ macro to print compiler info

1480c1:
Will print out the same thing if using gcc. Will print out the gcc compatible version plus extra information for clang.

`GCC 8.3.0` for gcc
`GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.5)`
`GCC 4.2.1 Compatible Clang 7.0.1 (tags/RELEASE_701/final)` for clang

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -76,7 +76,7 @@
 
 // Config Set Initial Count
 #define EB_SequenceControlSetPoolInitCount              3
-      
+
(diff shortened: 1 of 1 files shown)


tianjunwork:
@1480c1 Thanks for also cleaning up the trailing spaces. Which tool do you use? We may want to run it through the whole code base.

1480c1:
@tianjunwork I just use visual studio code. It automatically does that when I save a file

PR #140, opened 2019-03-08
Optimize /proc/cpuinfo parsing time on xeon-e5 with multiple cores

daijh:

PR #172, opened 2019-04-04
Fix a regression which increases init time

tianjunwork:


tianjunwork:
Duplicate with #140.

tianjunwork:
Using open() costs about 1s. This patch reduce it to 1.5ms.

PR #174, opened 2019-04-08
Remove SimpleApp

1480c1:


lijing0010:
why remove this simple app?

1480c1:
Similar to <https://github.com/OpenVisualCloud/SVT-AV1/pull/197>
and <https://github.com/OpenVisualCloud/SVT-AV1/issues/186>
It doesn't seem like a good idea to have to support two executables that essentially do the same thing, one just a stripped down version with different command line options.
Plus in the future, it will be hard to maintain parity and ensure that both work properly.

PR #193, opened 2019-05-08
add level constraint for multi tile and slice case

kelvinhu325:
add level constraint for multi tile and slice case

Signed-off-by: Hu, Kelvin <kelvin.hu@intel.com>


Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5228,6 +5230,18 @@ EB_ERRORTYPE ComputeProfileTierLevelInfo(
 
 	}
 
+    if(scsPtr->tileColumnCount > 1 || scsPtr->tileRowCount > 1) {
+        unsigned int levelIdx = 0;
+        const static general_level_idc[13] = {30, 60, 63, 90, 93, 120, 123, 150, 153, 156, 180, 183, 186};
```
lijing0010: why const static

PR #200, opened 2019-05-21
Switch CMake release builds to -O3

EwoutH:


tianjunwork:
Hi @EwoutH, "-D_FORTIFY_SOURCE=2 -O2" is recommended by our security team for product release. -O2 or higher is required by -D_FORTIFY_SOURCE=2. So I think -O3 will also work.
But I read online that:
"For many applications, -O2 is a good choice because the additional inlining and loop unrolling introduced by -O3 increases the instruction cache footprint, which ends up reducing performance."
Did you perform benchmark with -O3?

EwoutH:
On SVT-AV1 it improved performance with 2.5%. I will benchmark SVT-HEVC and SVT-VP9 tonight. Visual Studio also only accepts -O2, so that should also be fixed.

1480c1:
Can you remove O2/3 and fortify source from the regular cflags and add them to the release flags only? Some of my compilers complain that fortify source cannot be used with debug flags when building debug

tianjunwork:
> Can you remove O2/3 and fortify source from the regular cflags and add them to the release flags only? Some of my compilers complain that fortify source cannot be used with debug flags when building debug

I am interested to know which compilers?

1480c1:
Alpine GCC and clang

1480c1:
Those are the ones that come to mind immediately

1480c1:
I can try more when I get home, but those gave the warning when I was compiling svt-av1

1480c1:
from fedora
```
/usr/include/features.h:381:4: warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Wcpp]
 #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
```

1480c1:
It seems that a lot of other distros compilers don't care

tianjunwork:
Close. Included in #206.

PR #209, opened 2019-06-05
FFmpeg-Plugin: Switch to encode2 API and refine code

jsunintel:
Switched to encode2 API from send_frame/receive_packet API, for the latter has FPS being unstable per second issue. Also refined the code lines.

This patch has passed full cycle validation.


jsunintel:
@lijing0010 @hassount Please kindly review.

PR #211, opened 2019-06-09
Fix error of QP in RDOQ during 10-bit coding.

luofalei:
The `qp` here in function `DecoupledQuantizeInvQuantizeLoops()` was utilized for deriving weighted chroma distortion. It should be noted that the range of `qp` should be in `[0, 51]`.

However, the `qp` value was derived by adding the frame QP with `QP_BD_OFFSET` in 10-bit encoding in the EncDec process. This would cause the QP larger than `51` when the frame QP is large and cause reading overflow of the defined chroma weighting factor arrays.

In other modules for Mode-Decision, the weighted distortion was derived with the frame QP (without adding `QP_BD_OFFSET`),  So I suggest to apply this patch to fix the QP here  to avoid the above problem.


Source/Lib/Codec/EbFullLoop.c
```diff
@@ -525,6 +526,7 @@ void UnifiedQuantizeInvQuantize_R(
 			temporalLayerIndex < 3 ? EB_TRUE : EB_FALSE,
             (EB_U8)0,
             (EB_U16)qp,
+            bitDepth,
```
lijing0010: Do I have to set this to 8bit?

luofalei: Yes, it has been set to 8bit as follows.
https://github.com/OpenVisualCloud/SVT-HEVC/blob/882f926e02711d1825c62202e578c1d6e6839f43/Source/Lib/Codec/EbFullLoop.c#L488

lijing0010: thanks for catching this, it seems a pretty bad one for 10bit.

Issue #212, opened 2019-06-11
Library names conflict with SVT-VP9


ihalip:
I'm attempting to create a project using both SVT-HEVC and SVT-VP9 by building them in place, but each of them includes a set of libraries with the same name: C_DEFAULT, ASM_AVX2, etc.

Here's a sample CMake error message (there are more of them, for each of the duplicate names):
```
CMake Error at SVT-VP9/Source/Lib/C_DEFAULT/CMakeLists.txt:25 (add_library):
  add_library cannot create target "C_DEFAULT" because another target with
  the same name already exists.  The existing target is a static library
  created in source directory
  "/home/ihalip/Projects/repro/SVT-HEVC/Source/Lib/C_DEFAULT".  See
  documentation for policy CMP0002 for more details.
```

To reproduce: https://gist.github.com/ihalip/6a3668929b061aa32fa5feb9ffcb8778

An obvious fix would be to change the library names in their respective projects, because AFAICT there are large differences between these libraries in each project.

For consistency, I would suggest these libraries be renamed according to the project they are a part of, e.g. HEVC_C_DEFAULT and VP9_C_DEFAULT.

ihalip:
BTW, AVT-AV1 prefixes the names of these libs with `COMMON_`, so it has e.g. COMMON_C_DEFAULT. IMO, these should also be changed for consistency.

ihalip:
Here's a tentative change for all 3 projects: https://gist.github.com/ihalip/57e6fd4548e66501bde247ee6c3a146e

This introduces a `SVT_COMMON_LIBRARY_PREFIX` variable (set to either AV1_, HEVC_, VP9_), and the names of the common libraries are prefixed with this value. Here's a list of generated targets:

```
$ make help | grep -E "(ASM|DEFAULT)"
... AV1_C_DEFAULT
... AV1_ASM_SSE2
... AV1_ASM_SSSE3
... AV1_ASM_SSE4_1
... AV1_ASM_AVX2
... HEVC_C_DEFAULT
... HEVC_ASM_SSE2
... HEVC_ASM_SSSE3
... HEVC_ASM_SSE4_1
... HEVC_ASM_AVX2
... VP9_C_DEFAULT
... VP9_ASM_SSE2
... VP9_ASM_SSSE3
... VP9_ASM_SSE4_1
... VP9_ASM_AVX2
```

1480c1:
[...]

PR #215, opened 2019-06-12
cmake: use a prefix for the common libraries

ihalip:
Should fix [SVT-HEVC/212](https://github.com/OpenVisualCloud/SVT-HEVC/issues/212).


1480c1:
I've added you changes to https://github.com/OpenVisualCloud/SVT-HEVC/pull/206, would that be fine?

Issue #118, opened 2019-02-26
Unable to compile with MinGW-w64


filler56789:
When running "build.sh", CMake returns an error message and stops.
And no, using MSVC is not a proper solution — actually it's not a solution at all.

1480c1:
Indeed, https://travis-ci.com/1480c1/SVT-VP9/jobs/180668415

1480c1:
Segmentation fault when using mingw-gcc. Seems to work with mingw clang though

filler56789:
Well, the least that you could and should do is release a Windows binary. And drop the CMake nonsense entirely. Nobody should be forced to use Visual Studio when the goal is create a command-line application.

1480c1:
Appveyor will be implemented once they transfer the repo to OpenVisualCloud from Intel. I do want the program to be able to work with mingw though.

tianjunwork:
@filler56789 and @1480c1 for posting the question. I agree that another build option should be provided on Windows. Since MinGW-clang works, could you work on a PR to update the README for the instruction to build with clang on Windows? Thank you!

1480c1:
> @filler56789 and @1480c1 for posting the question. I agree that another build option should be provided on Windows. Since MinGW-clang works, could you work on a PR to update the README for the instruction to build with clang on Windows? Thank you!

Sure, however I would probably wait until the cmakefiles are edited since if compiled as is, it will only compile with -march=x86_64 and no other flags since the cmakefiles only list gnu GCC and Intel and not clang. The fPIE and fPIC flags will need to be separated with if not mingw and the optimize unroll loops will need to use   if clang pragma since clang doesn't support __attribute optimize unroll loop thing. Might submit a pr later if I can make sure it works.

1480c1:
It seems I have lied on the mingw-clang being able be usable. It will compile without error and there is no segmentation faults with the resulting file. However, the output video file looks like: 
![image](https://user-images.githubusercontent.com/8345542/53675145-44784780-3c59-11e9-8ef1-fd5c52fd8d18.png)
Which I am pretty sure is not intended.

filler56789:
Yet another problem:
please consider replacing "#ifdef _MSC_VER" with "#if defined(_WIN32) || defined(_WIN64)" in EbAppMain.c,

otherwise stdin and stdout will NOT work with the MinGW-w64 builds.
By the way, stdin apparently does not work with the INTEL compiler builds either.

[...]

PR #228, opened 2019-06-26
CMake: apply -fstack-protector-strong if not mingw

1480c1:
Don't apply stack protector if mingw since not all mingw-w64 gcc's are compiled with stack-protector included

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,6 @@ set(flags_to_test
     -Wextra
     -Wformat
     -Wformat-security
-    -fstack-protector-strong
     -fvisibility=hidden
     -fPIE
     -fPIC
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -96,6 +95,8 @@ else()
     #for Mingw64 support, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782
     if(MINGW)
         list(APPEND flags_to_test -fno-asynchronous-unwind-tables)
+    else()
+        list(APPEND flags_to_test -fstack-protector-strong)
```
tianjunwork: why it doesn't apply also to MINGW?
Is there some performance sacrifice?

1480c1: As mentioned in https://github.com/OpenVisualCloud/SVT-HEVC/issues/118, some builds of mingw-w64 toolchain does not have the required stuff stack protector. Although some do, but it would be hard to properly detect it in cmake due to the errors appearing during link time

tianjunwork: Do you have an estimation of performance overhead of using -fstack-protector-strong?

1480c1: As far as I know, stack protector is more of a security thing rather than performance. As far as I know, for systems other than MINGW, the performance should be the same. It is only for when compiling using MINGW, that stack protector is not applied

tianjunwork: I did some test. Didn't see performance impact. Thx.

PR #244, opened 2019-07-11
Fix reading overflow of `skipFlagBits[]` and typos in `EbRateDistortionCost.c`.

luofalei:
(1)  The loop-up table `skipFlagBits` was fixed.

In HEVC, there are 3 contexts for `skip_flag` and there are 6 values in the look-up table as follows.
```
static const EB_U32 skipFlagBits[6] = { 17878, 62157, 86651, 54723, 14816, 8254 };
```
This may lead to overflow in bit estimation of skip_flag in function `InterFastCostPsliceOpt()` and `InterFastCostBsliceOpt()`.
```
	// Rate
	EB_U64 mergeSkiplumaRate;
	mergeSkiplumaRate = skipFlagBits[4 + cuPtr->skipFlagContext];
```

(2) some typos in this file was fixed:
- Esimate => Estimate;
- Assigne => Assign;
- decisoion => decision;
- condidate => candidate;

### DIFF
--- a/Source/Lib/Codec/EbRateDistortionCost.c
+++ b/Source/Lib/Codec/EbRateDistortionCost.c
@@ -11,7 +11,7 @@
 
 #include "EbModeDecisionConfiguration.h"
 
-static const EB_U32 skipFlagBits[6] = { 17878, 62157, 86651, 54723, 14816, 8254 };
+static const EB_U32 skipFlagBits[NUMBER_OF_SKIP_FLAG_CASES] = { 17878, 62157, 86651, 54723, 14816, 8254 };
 static const EB_U32 mergeIndexBits[6] = { 10350, 109741, 142509, 175277, 175277 };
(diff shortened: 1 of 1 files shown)


lijing0010:
@luofalei ,thanks for catching this.
btw: @hassount  For the following one, shall we change it from 6 to 5, why "mergeIndexBits[**6**]" here?

static const EB_U32 mergeIndexBits[6] = { 10350, 109741, 142509, 175277, 175277 };

luofalei:
> @luofalei ,thanks for catching this.
> btw: @hassount For the following one, shall we change it from 6 to 5, why "mergeIndexBits[**6**]" here?
> 
> static const EB_U32 mergeIndexBits[6] = { 10350, 109741, 142509, 175277, 175277 };

There are at most `5` merge candidates in `HEVC`, so I suggest to change it from 6 to 5, which is equal to `NUMBER_OF_MERGE_INDEX_CASES` in `MdRateEstimationContext_s`.

PR #245, opened 2019-07-11
Nalu file processing

intelmark:
NALU file processing

1) Fixed strol related Klocwork issues in EbAppProcessCmd.c by breaking
out EB_STRTOK macro onto separate line and then checking for NULL.

2) Correctly buffer size checking in BaseDecodeFunction.  Using a
valid NALU file, exising code was returning EB_ErrorBadParameter because
it was checking the buffer range based on sizeof(U8 pointer). Fixed by
adding base64DecodeLength parameter

3)  Adding (uin32_t) casts to eliminate compile warning message in
Windows build.


intelmark:
will recreate the PR..

PR #246, opened 2019-07-11
Y4m ctype

intelmark:
Adding ctype.h include

Adding <ctype.h> include in EbAppInputy4m.c to resolve compiler
warning about 'isalnum' undefined.  Using isalnum for checking
for alphanumeric characters (part of previous Klocwork fix).


tianjunwork:
Hi @intelmark , why there are many duplication with #245 ?

PR #247, opened 2019-07-11
NALU file processing

intelmark:
1) Split EB_STRTOK macro into 2 steps to check for NULL  return (several
Klocwork issues)

2) Added (uint32_t) cast to eliminate compiler warning

3) Decoding buffer size check was failing because of sizeof(decodedString)
check.
Added base64DecodeLength parameter to BaseDecodeFunction call and use that
parameter instead for checking buffer length.

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -3137,7 +3137,7 @@ EB_API EB_ERRORTYPE EbH265EncEosNal(
 }
 
 /* charSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/" */
-static EB_ERRORTYPE BaseDecodeFunction(EB_U8* encodedString, EB_U32 base64EncodeLength, EB_U8* decodedString )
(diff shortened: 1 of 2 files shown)


Source/Lib/Codec/EbEncHandle.c
```diff
@@ -3202,15 +3203,18 @@ static EB_ERRORTYPE ParseSeiMetaData(
...
     }
 
     base64Encode = src->naluBase64Encode;
     base64EncodeLength = (uint32_t)strlen((char*)base64Encode);
-    EB_MALLOC(EB_U8*, base64Decode, (base64EncodeLength / 4) * 3, EB_N_PTR);
-    return_error = BaseDecodeFunction(base64Encode, base64EncodeLength, base64Decode);
+    base64DecodeLength = (base64EncodeLength / 4) * 3;
+    EB_MALLOC(EB_U8*, base64Decode, base64DecodeLength, EB_N_PTR);
+    
+    return_error = BaseDecodeFunction(base64Encode, base64EncodeLength, base64Decode, base64DecodeLength);
```
tianjunwork: good catch!

PR #257, opened 2019-07-17
STYLE.md

intelmark:
Adapted STYLE.md file from AV1 project and modified for HEVC project.
Added mention of STYLE.md in README.md
Removed some unneeded trailing slashes in README.md

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -3,7 +3,7 @@
 
 This style guide should be used for all SVT-HEVC code submissions
 
-Tabs vs Spaces
+Tabs vs Spaces\
 **No tabs,** only spaces; 4-space indentation
 
 Be aware that some tools might add tabs when auto aligning the code, please check your commits with a diff tool for tabs.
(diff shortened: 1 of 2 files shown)


STYLE.md
```diff
@@ -0,0 +1,152 @@
...
+
+Note: For macOS and BSD related distros, you may need to use `sed -i ''` inplace due to differences with GNU sed.
+
+### For Powershell/pwsh
+
+``` Powershell
+ls -Recurse -File -Filter *.c | ForEach-Object{$(Get-Content $_.FullName | Foreach {Write-Output "$($_.TrimEnd().Replace("`t","    "))`n"}) | Set-Content -NoNewline -Encoding utf8 $_.FullName}
+```
+
+Where `-Filter *.c` has your extention/filename(s).
```
tianjunwork: You need \ here, otherwise there is no newline

tianjunwork: Pls compare with the one from AV1 to see where \ is needed.

STYLE.md
```diff
@@ -0,0 +1,152 @@
+
+# Code Style
+
+This style guide should be used for all SVT-HEVC code submissions
+
+Tabs vs Spaces
```
tianjunwork: Add \ after line 6

Issue #269, opened 2019-07-23
Encode output bitstream has obvious garbage with P444 480/720/1080 8bit/10bit YUV input in encMode 0-2


liweki:
During run SVT-HEVC test with Commit ID 94996f6, found the encode output bitstream has green screen or obvious garbage with **P444 480/720/1080 8bit YUV input in encMode 0-2** & **P422/P444 10bit 480 YUV input in encMode 0-9**.
A little encoded 4k bitstreams also has this issue.

Pass on pervious commit ID 7ef5fc8.

**Test command line we used:**
**8bit P444:**
- encoders/SvtHevcEncApp -encMode 2  -w 1920  -h 1080  -i yuvs/WITCHER3_007_1920x1080_8bit_60Hz_P444.yuv  -q 37  -fps 60  -b bitstreams/M2_T1_WITCHER3_007_1920x1080_8bit_60Hz_P444_Q37.bin   -n 600  -intra-period 55  -bit-depth 8 -color-format 3
- encoders/SvtHevcEncApp -encMode 0  -w 3840  -h 2160  -i yuvs/Netflix_Crosswalk_3840x2160_10bit_60Hz_P444.yuv  -q 22  -fps 60  -b bitstreams/svt-hevc_M0_T1_Netflix_Crosswalk_3840x2160_10bit_60Hz_P444_Q22.bin   -n 600  -intra-period 55  -bit-depth 10 -color-format 3

**10bit P444:**
- encoders/p/SvtHevcEncApp -encMode 0 -w 864  -h 480  -i yuvs/ParkJoy_864x480_10bit_50Hz_P444.yuv  -q 32  -fps 50  -b bitstreams/M0_T1_ParkJoy_864x480_10bit_50Hz_P444_Q32.bin   -n 600  -intra-period 47  -bit-depth 10 -color-format 3

Attach one frame as reference:

![garbage](https://user-images.githubusercontent.com/46477459/61684341-c9204b00-ad4b-11e9-9594-084b3a86cc68.png)

tianjunwork:
Hi @liweki, Thx again pointing out the regression.

kelvinhu325:
I tried it, it's proved the commit https://github.com/OpenVisualCloud/SVT-HEVC/commit/94996f6d5041c7830e34f95be4706b151f8624f8 causes this artifacts.

intelmark:
[...]

PR #275, opened 2019-07-24
Fix the deltaQP issue for intra4x4 with YUV444

lijing0010:


lijing0010:
Fix bug #269

Issue #281, opened 2019-07-26
[SVT-HEVC]Some test found the md5 value of encoded stream is differ but have same bitstream size between AsmType enable and disable.


GuCaixuan:
[Test environment]
OS: Ubuntu 18.04.2 LTS
SVT CommitID: 1a9474a

[Issue Description]
During run SVT-HEVC test with Commit ID 1a9474a, found the md5 value of the encoded streams is different between AsmType enable and disable when encoding with some special combination of encoding parameters.
For example:
1920x1080_P422 input, encMode 0, qp 41
1920x1080_P444 input, encMode 4, qp 33
7680x3840_P420 input, encMode 6, qp 35

Test command line we used:
encoders/SvtHevcEncApp -i /home/media/yuvs/SVT/yuvs/Fallout4_1920x1080_8bit_60Hz_P422.yuv  -b asm_test_M0_Fallout4_1920x1080_8bit_60Hz_P422_Q41_OQ_0.265 -w 1920  -h 1080  -bit-depth 8  -fps 60  -intra-period 55  -n 20  -encMode 0  -tune 1  -asm 0  -compressed-ten-bit-format 0  -q 41  -color-format 2  -rc 0 
encoders/SvtHevcEncApp -i /home/media/yuvs/SVT/yuvs/Fallout4_1920x1080_8bit_60Hz_P422.yuv  -b asm_test_M0_Fallout4_1920x1080_8bit_60Hz_P422_Q41_OQ_1.265 -w 1920  -h 1080  -bit-depth 8  -fps 60  -intra-period 55  -n 20  -encMode 0  -tune 1  -asm 1  -compressed-ten-bit-format 0  -q 41  -color-format 2  -rc 0

encoders/SvtHevcEncApp -i /home/media/yuvs/SVT/yuvs/Fallout4_1920x1080_8bit_60Hz_P444.yuv  -b asm_test_M4_Fallout4_1920x1080_8bit_60Hz_P444_Q33_OQ_0.265 -w 1920  -h 1080  -bit-depth 8  -fps 60  -intra-period 55  -n 20  -encMode 4  -tune 1  -asm 0  -compressed-ten-bit-format 0  -q 33  -color-format 3  -rc 0 
[...]

PR #292, opened 2019-07-30
[ASM_SSE2] fixed the output mismatching issue with ASM enabled and disabled (Issue #281).

Austin-Hu:
When gathering SAO statistics for each LCU, the invalid samples
beyond the right edge of LCU shouldn't be counted, although the
use SSE2 intrinsics can only support __m128i data type. So fixed
the issue for SAO BO counting, as the invalid samples are always
0 and would be counted into BO index 0 containing sample value 0~7.

As for EO, all of the C_DEFAULT implementations
(such as GatherSaoStatisticsLcuLossy_62x62) will shifts edge index
2 and 3, with 3 and 4 separately. And the SSE2 implementation
(countEdge) doesn't handle edge index. So the encoded outputs
wouldn't have bit diff, although the EO diffs and counts of edge
index 4 have some gap, between ASM enabled and disabled.

### DIFF
--- a/Source/Lib/ASM_SSE2/EbSaoStatistics_Intrinsic_SSE2.c
+++ b/Source/Lib/ASM_SSE2/EbSaoStatistics_Intrinsic_SSE2.c
@@ -59,17 +59,39 @@ static void countEdge(__m128i *eoDiff, __m128i *eoCount, EB_BYTE ptr, EB_S32 off
 }
 
 
-static void countBand(EB_S32 *boDiff, EB_U16 *boCount, EB_S32 cat01, EB_S32 diff01)
+static void countBand(EB_S32 *boDiff, EB_U16 *boCount, EB_S32 cat01, EB_S32 diff01, EB_U8 *validSample)
 {
     EB_U8 cat0 = (EB_U8)cat01;
     EB_U8 cat1 = (EB_U8)(cat01 >> 8);
     EB_S8 diff0 = (EB_S8)diff01;
     EB_S8 diff1 = (EB_S8)(diff01 >> 8);
 
-    boCount[cat0]++;
-    boCount[cat1]++;
-    boDiff[cat0] += diff0;
-    boDiff[cat1] += diff1;
+    if (validSample[0])
+    {
+        boCount[cat0]++;
+        boDiff[cat0] += diff0;
+    }
+
+    if (validSample[1])
+    {
+        boCount[cat1]++;
+        boDiff[cat1] += diff1;
+    }
+}
+
+static void updateValidSamples(EB_U8 *validSample, EB_U8 size, EB_S32 colCount, EB_S32 colCountDiv2)
+{
+    if (!validSample)
+        return;
+
+    EB_MEMSET(validSample, 1, size);
+    if (colCountDiv2 == 0)
+    {
+        validSample[0] = (colCount & 1) ? 1 : 0;
+        validSample[1] = 0;
+    }
+    else if (colCountDiv2 < 0)
+        EB_MEMSET(validSample, 0, size);
 }
 
 EB_EXTERN EB_ERRORTYPE GatherSaoStatisticsLcu_BT_SSE2(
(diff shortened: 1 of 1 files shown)


tianjunwork:
Hi Austin:
There are build errors on Windows:
(ClCompile target) -> 
  C:\projects\svt-hevc-lopps\Source\Lib\ASM_SSE2\EbSaoStatistics_Intrinsic_SSE2.c(96,38): error C2057:  expected constant expression [C:\projects\svt-hevc-lopps\Build\Source\Lib\ASM_SSE2\HEVC_ASM_SSE2.vcxproj]
  C:\projects\svt-hevc-lopps\Source\Lib\ASM_SSE2\EbSaoStatistics_Intrinsic_SSE2.c(96,38): error C2466:  cannot allocate an array of constant size 0 [C:\projects\svt-hevc-lopps\Build\Source\Lib\ASM_SSE2\HEVC_ASM_SSE2.vcxproj]
  C:\projects\svt-hevc-lopps\Source\Lib\ASM_SSE2\EbSaoStatistics_Intrinsic_SSE2.c(96,39): error C2133:  'validSample': unknown size [C:\projects\svt-hevc-lopps\Build\Source\Lib\ASM_SSE2\HEVC_ASM_SSE2.vcxproj]

ttrigui:
The fix looks legit however I'm not sure how this will impact the speed. Probably a better solution would be to call countBand only when needed.
a simple solution can be something similar to this:
  if( 7 < colCountDiv2)
            {
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 0), _mm_extract_epi16(diff, 0), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 1), _mm_extract_epi16(diff, 1), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 2), _mm_extract_epi16(diff, 2), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 3), _mm_extract_epi16(diff, 3), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 4), _mm_extract_epi16(diff, 4), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 5), _mm_extract_epi16(diff, 5), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 6), _mm_extract_epi16(diff, 6), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 7), _mm_extract_epi16(diff, 7), validSample);
            }
            else if (6 < colCountDiv2)
            {
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 0), _mm_extract_epi16(diff, 0), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 1), _mm_extract_epi16(diff, 1), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 2), _mm_extract_epi16(diff, 2), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 3), _mm_extract_epi16(diff, 3), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 4), _mm_extract_epi16(diff, 4), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 5), _mm_extract_epi16(diff, 5), validSample);
                countBand(boDiff, boCount, _mm_extract_epi16(cat, 6), _mm_extract_epi16(diff, 6), validSample);
            }
else if (5 < colCountDiv2)
{........

1480c1:
@ttrigui, please try to use code blocks, they help differentiate between what you are trying to represent as code and the rest, plus it helps with keeping formatting etc

https://help.github.com/en/articles/creating-and-highlighting-code-blocks

ttrigui:
@1480c1 thank you. I hope now it looks better.

Austin-Hu:
Hi @ttrigui ,

Either way is OK, although the ideal implementation should loop calling countBand but we can not use variable as the 2nd argument of _mm_extract_epi16() as I mentioned in the comments. I was just thinking to make the GatherSaoStatisticsLcu_BT_SSE2() function as shorter as possible :p. Thanks!

ttrigui:
@Austin-Hu Correct the second argument to _mm_extract_epi16() cannot be variable which is why I suggeted a solution with an if condition. I still believe that the latter is more efficient as it has less checks and no extra code.

tianjunwork:
Thank you @ttrigui for reviewing the PR. Both implementation is good from different perspective. I think a more maintainable/readable, clearer coding style is worth the sacrifice of a few if cycle. I will merge the PR for closing the bug. Could @Austin-Hu do a benchmark later to compare the performance? Thank you both!

Austin-Hu:
BTW, I’ve verified the 3 reported cases (in Issue [#281](https://github.com/OpenVisualCloud/SVT-HEVC/issues/281)) with several times, with and without this PR applied, and there wasn’t any performance gap (normally +/- 2 FPS) between them. And I would notice any performance regression once reported (from validation team). Thanks!

PR #295, opened 2019-08-01
PRINT_HEADER Preprocessor

intelmark:
Undefined PRINT_HEADER to prevent the display of y4m header values

### DIFF
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -8,7 +8,9 @@
 #define YUV4MPEG2_IND_SIZE 9
 
 // Define PRINT_HEADER to log y4m header contents
-//#define PRINT_HEADER
+#ifndef NDEBUG
+#define PRINT_HEADER
+#endif
 
 #define CHROMA_MAX 4
 
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -7,11 +7,6 @@
 #define YFM_HEADER_MAX 80
 #define YUV4MPEG2_IND_SIZE 9
 
-// Define PRINT_HEADER to log y4m header contents
-#ifndef NDEBUG
-#define PRINT_HEADER
-#endif
-
 #define CHROMA_MAX 4
 
 #include "EbAppConfig.h"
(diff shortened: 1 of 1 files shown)


1480c1:
Could this be gated with
``` c
#ifndef NDEBUG
#define PRINT_HEADER 0
#endif
```
?
That way it will still show in debug builds?

intelmark:
Not  a bad idea..  Also it’s the presence of the symbol (not the value) that triggers the logging:


#ifndef NDEBUG

#define PRINT_HEADER

#endif

Source/App/EbAppInputy4m.c
```diff
@@ -6,7 +6,12 @@
 #include "EbAppInputy4m.h"
 #define YFM_HEADER_MAX 80
 #define YUV4MPEG2_IND_SIZE 9
-#define PRINT_HEADER 0
+
+// Define PRINT_HEADER to log y4m header contents
+#ifndef NDEBUG
+#define PRINT_HEADER
+#endif
+
```
tianjunwork: Hi, even a debug build, we don't really need to print these logs. They are not something special that we need to pay attention in debug session. Please turn if off too for debug build.

1480c1: I would argue that it would have a value in debug mode when you are trying to confirm what the encoder sees when reading y4m files to make sure there isn't a problem with the y4m video nor in the y4m code

tianjunwork: If for this reason, there will so many more logs to print. 
To validate y4m, ffprobe can be used. 
And it is easy to debug those printed values in VS if anything wrong in y4m code.

1480c1: I guess

intelmark: I will turn off in debug builds.  Note this was in the AV1 implementation I ported this feature from.

tianjunwork: Thx Mark. AV1 is still under dev. Clean up is needed.

intelmark: Just to be clear...should I remove the PRINT_HEADER feature all together, or just disable in debug/release builds (e.g. #undef PRINT_HEADER). imho it's useful when debugging

tianjunwork: Pls remove this line: #define PRINT_HEADER 0

1480c1:
Would it be better to clean up the associated code if print header is going to removed?

tianjunwork:
I am fine keep them there. The developer can add "#define PRINT_HEADER" to enable the print.

PR #302, opened 2019-08-06
Cleanup: Convert tabs to spaces and remove trailing spaces

1480c1:
WIll the patches be okay with the conversion of whitespaces? Mainly about the hashes

Also modified STYLE.md's sed to exclude .git, build,Bin, and png and added a command to append newlines to files if it doesn't have it. I checked the encoder guide and the pictures seem to not be corrupt


1480c1:
The style job now passes for tabs, carriage returns, and spaces and final newlines

Should we enforce that now?

tianjunwork:
@1480c1 Thank you for taking the effort to fix the coding style. We intentionally postpone coding style clean up to avoid rebasing of #70. Once it is merge, could you perform clean up again? Thanks.

1480c1:
Sure!

intelmark:
Can this be updated as it is the remaining PR item for the 1.4 release and now is a good time to make these changes?  There was another PR for cleanup (226) that is a bit older and can be withdrawn.

PR #303, opened 2019-08-06
Support both dynamic/static builds

### LINKED ISSUE #255: when to support building both static/dynamic libs
when to support building both static/dynamic libs?

1480c1:
Closes #255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,10 @@ notifications:
   webhooks:
     - https://coveralls.io/webhook
 env:
-  - build_type=release
+  global: build_type=release
+  matrix:
+    - CMAKE_EFLAGS="-DBUILD_SHARED_LIBS=ON"
+    - CMAKE_EFLAGS="-DBUILD_SHARED_LIBS=OFF"
 stages:
   - name: style
   - name: test
@@ -110,7 +113,7 @@ matrix:
         #- git am $TRAVIS_BUILD_DIR/ffmpeg_plugin/0002-doc-Add-libsvt_hevc-encoder-docs.patch
         - export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/lib/pkgconfig
         - "sudo chown -R travis: $HOME/.ccache"
-        - ./configure --enable-libsvthevc || less ffbuild/config.log
+        - ./configure --enable-libsvthevc || tail ffbuild/config.log
         - make --quiet -j$(nproc) && sudo make install
     # GStreamer interation build
     - name: GStreamer patch
--- a/Build/linux/build.sh
+++ b/Build/linux/build.sh
@@ -3,69 +3,506 @@
 # Copyright(c) 2018 Intel Corporation
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 
-function clean {
-	rm -R -f debug
-	rm -R -f release
-    rm -R -f ../../Bin/Debug
-    rm -R -f ../../Bin/Release
+# Fails the script if any of the commands error (Other than if and some others)
+set -e
+
+# Allow use for coloring (as long as it's within the lines)
+if [ -n "$(tput colors 2> /dev/null)" ] && [ "$(tput colors 2> /dev/null)" -ge 8 ]; then
+    bold=$(tput bold)
+    blue=$(tput setaf 12)
+    yellow=$(tput setaf 11)
+    purple=$(tput setaf 13)
+    green=$(tput setaf 2)
+    red=$(tput setaf 1)
+    reset=$(tput sgr0)
+fi
+
+# Function to print to stderr or stdout while stripping preceding tabs(spaces)
+# If tabs are desired, use cat
+print_message() {
+    if [ "$1" = "stdout" ] || [ "$1" = "-" ]; then
+        shift
+        printf "%b\\n" "${@:-Unknown Message}" | sed -e 's/^[ \t]*//'
+    else
+        printf "%b\\n" "${@:-Unknown Message}" | sed -e 's/^[ \t]*//' >&2
+    fi
 }
 
-function debug {
-	mkdir -p debug
-	mkdir -p ../../Bin/Debug
-	cd debug
-	PATH=$PATH:/usr/local/bin/
-	cmake ../../.. 					\
-		-DCMAKE_BUILD_TYPE=Debug			\
-		-DCMAKE_C_COMPILER=$CMAKE_COMPILER		\
-		-DCMAKE_ASM_NASM_COMPILER=$CMAKE_ASSEMBLER	\
-		-DCMAKE_AR=`which $AR_COMPILER`			\
-		-DCMAKE_RANLIB=`which $RANLIB_COMPILER`	\
-
-	# Compile the App
-	make -j SvtHevcEncApp
-	cd ..
+# Convenient function to exit with message
+die() {
+    if [ -z "$*" ]; then
+        print_message "${red}Unknown error${reset}"
+    else
+        print_message "$@"
+    fi
+    print_message "The script will now exit."
+    exit 1
 }
 
-function release {
-	mkdir -p release
-	mkdir -p ../../Bin/Release
-	cd release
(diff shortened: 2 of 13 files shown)


.travis.yml
```diff
@@ -26,7 +26,10 @@ notifications:
   webhooks:
     - https://coveralls.io/webhook
 env:
-  - build_type=release
+  global: build_type=release
+  matrix:
+    - CMAKE_EFLAGS="-DBUILD_SHARED_LIBS=ON"
+    - CMAKE_EFLAGS="-DBUILD_SHARED_LIBS=OFF"
```
1480c1: I can remove this is not wanted

CMakeLists.txt
```diff
@@ -85,7 +85,6 @@ set(flags_to_test
     -Wextra
     -Wformat
     -Wformat-security
-    -fvisibility=hidden
```
1480c1: I removed this since it was causing issues with building so with functions being hidden

gstreamer-plugin/git-patch/0001-svthevcenc-Add-svthevc-encoder-element.patch
```diff
@@ -1868,10 +1872,10 @@ index 0000000..53be342
 +    plugin_init, VERSION, "GPL", GST_PACKAGE_NAME, GST_PACKAGE_ORIGIN)
 diff --git a/ext/svthevcenc/gstsvthevcenc.h b/ext/svthevcenc/gstsvthevcenc.h
 new file mode 100644
-index 0000000..be1d492
+index 000000000..ad54b9c83
 --- /dev/null
 +++ b/ext/svthevcenc/gstsvthevcenc.h
-@@ -0,0 +1,118 @@
+@@ -0,0 +1,116 @@
```
tianjunwork: Regen patch file can be avoid by modifying the file directly. 
Only change line 1874 and remove line 1901,1902.
Similar with ffmpeg

gstreamer-plugin/git-patch/0001-svthevcenc-Add-svthevc-encoder-element.patch
```diff
@@ -1898,8 +1898,6 @@ index 0000000..be1d492
 +#include <gst/video/video.h>
 +#include <gst/video/gstvideoencoder.h>
 +
```
tianjunwork: You do need to change line 1874: reduce 1,118 by 2. Otherwise it doesn't apply.
same for ffmpeg

lijing0010:
it depends on ICC by default?
jingli@media-lijing-3 ~/merge/static/SVT-HEVC/Bin/release$ ldd libSvtHevcEnc.so
        linux-vdso.so.1 =>  (0x00007ffdd9df3000)
        libimf.so => not found
        libsvml.so => not found
        libirng.so => not found
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb95dd6c000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb95db56000)
        libintlc.so.5 => not found
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb95d78c000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fb95d588000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb95e668000)

and I got a lot of message like "remark #11074: Inlining inhibited by limit max-size"

tianjunwork:
Still wip, not ready to review yet.

1480c1:
The script tries to find icc if no CC is passed, as that was the default in the previous versions of the build script

lijing0010:
> The script tries to find icc if no CC is passed, as that was the default in the previous versions of the build script

I guess I have ICC installed, but didn't add it in the PATH.

Build/linux/build.sh
```diff
@@ -3,69 +3,506 @@
...
+Usage: $THIS_SCRIPT [OPTION] ... -- [OPTIONS FOR CMAKE]
+-a, --all, all          Builds release and debug
+    --asm, asm=*        Set assembly compiler [$ASM]
+-b, --bindir, bindir=*  Directory to install binaries
+    --cc, cc=*          Set C compiler [$CC]
+    --cxx, cxx=*        Set CXX compiler [$CXX]
+    --clean, clean      Remove build and Bin folders
+    --debug, debug      Build debug
+    --shared, shared    Build shared libs
+-x, --static, static    Don't build shared libs
```
tianjunwork: "Build static lib"

Build/linux/build.sh
```diff
@@ -3,69 +3,506 @@
...
+
+# Help message
+echo_help() {
+    cat << EOF
+Usage: $THIS_SCRIPT [OPTION] ... -- [OPTIONS FOR CMAKE]
+-a, --all, all          Builds release and debug
+    --asm, asm=*        Set assembly compiler [$ASM]
+-b, --bindir, bindir=*  Directory to install binaries
+    --cc, cc=*          Set C compiler [$CC]
+    --cxx, cxx=*        Set CXX compiler [$CXX]
```
tianjunwork: do we really support cxx?

Build/linux/build.sh
```diff
@@ -3,69 +3,506 @@
...
+    --clean, clean      Remove build and Bin folders
+    --debug, debug      Build debug
+    --shared, shared    Build shared libs
+-x, --static, static    Don't build shared libs
+-g, --gen, gen=*        Set CMake generator
+-i, --install, install  Install build [Default release]
+-j, --jobs, jobs=*      Set number of jobs for make/CMake [$jobs]
```
tianjunwork: Keep this file as simple as possible. 
1. make it easier to maintain
2. straightforward to user

Let consider remove un-related/not used ones?
-ams
-bindir
-cxx
-gen
-jobs
-prefix
-toolchain
-verbose

You can hardcore some of them instead of having the option exposed.

Build/linux/build.sh
1480c1: Done

Build/linux/build.sh
1480c1: I don't think so, I can remove

Build/linux/build.sh
1480c1: Sure, although I would like to keep at least prefix and gen so that way the user can set the install prefix to be something other than /usr/local and use Ninja or some other generator if they prefer

1480c1:
Readme probably does need an update since it's showing the commands themselves (cmake make etc) instead of using build.sh, unless that's okay

Build/linux/build.sh
tianjunwork: SvtHevcEnc.pc has prefix=/usr/local which is used by gst plugin. Let's just use the defual intall dir, in case user mess up the plugin without knowing it.

1480c1: The pkg-config is auto generated based on cmake, so if prefix is passed as `./build.sh --prefix /usr`, the script will pass `-DCMAKE_INSTALL_PREFIX=/usr` to cmake, which in turn will generate

``` bash
prefix=/usr
includedir= ${prefix}/include/svt-hevc
...
```

tianjunwork: thx, let me try

1480c1: Right now doing final cleanup of the script for any dangling variables

tianjunwork: Does using Ninja tested?

1480c1: I think I did test at least the SVT-AV1, I can test right now

1480c1: It work, although since I remove `--gen`, it can be used by `./build.sh ... -- -G"Ninja"`

tianjunwork:
Thx for resolving the conflict.

1480c1:
Right now, the script doesn't work in alpine due to their lack of bash array support, going to test on centos

tianjunwork:
may be document the usage of build.sh in README?
the current instruction only shows how to build shared.

Build/linux/build.sh
tianjunwork: thx

1480c1:
Added some basic documentation for build.sh, please check if the wording makes sense.

PR #310, opened 2019-08-12
CMake: Move version value managing to top of CMakeList.txt

Yeongjin-Jeong:
Using the cmake environment variable as the version value simplifies versioning the pkg-config file or API header.

And the version value of the pkg-config file and API header has not changed in the latest release tag, so it was bumped up version to 1.4.0.


CMakeLists.txt
```diff
@@ -31,8 +31,14 @@ if(NOT DEFINED CMAKE_INSTALL_PREFIX OR CMAKE_INSTALL_PREFIX STREQUAL "")
 endif()
 message(STATUS "CMAKE_INSTALL_PREFIX = ${CMAKE_INSTALL_PREFIX}")
 
+set(SVT_VERSION_MAJOR           "1")
+set(SVT_VERSION_MINOR           "4")
+set(SVT_VERSION_PATCHLEVEL      "0")
```
1480c1:
```suggestion
set(SVT_HEVC_VERSION_PATCHLEVEL      "0")
```

Change to SVT_HEVC to separate out from the other SVT encoders?

Yeongjin-Jeong: Does that mean only cmake variables ? If that is right, I applied it :)

1480c1: Yes

tianjunwork:
Hi @Yeongjin-Jeong, thanks for your contribution. Thanks @1480c1 for helping code review.

Issue #305, opened 2019-08-08
when encoding with codeVpsSpsPps enabled, encoded packets are not output


Yeongjin-Jeong:
I used the gstreamer svthevcenc plugin with the latest SVT-HEVC, and found that the encoded packets are not output from the encoder.

After initializing the encoder with codeVpsSpsPps enabled, It seems to be caused by using the EbH265EncStreamHeader API before receiving encoded packets. And the same happens with the ffmpeg plugin or SvtHevcEncApp, if codeVpsSpsPps is enabled.

This occurs after patch https://github.com/OpenVisualCloud/SVT-HEVC/pull/155 is merged. Has the usage of EbH265EncStreamHeader API or codeVpsSpsPps flag changed? Or is there something wrong in my build environment?

Here is the lscpu output:
```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                16
On-line CPU(s) list:   0-15
Thread(s) per core:    2
Core(s) per socket:    8
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 63
Model name:            Intel(R) Xeon(R) CPU E5-2630 v3 @ 2.40GHz
Stepping:              2
CPU MHz:               1394.019
CPU max MHz:           3200.0000
CPU min MHz:           1200.0000
BogoMIPS:              4788.65
Virtualisation:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              20480K
NUMA node0 CPU(s):     0-15
[...]

PR #264, opened 2019-07-20
Optimize Copy of SequenceControlSet and initial value assignment

luofalei:
1) In function `CopySequenceControlSet()`, paramters were assigned one by one. This would introde bugs if some parameters were not set. I followed the assignements here and copied the beging part started with `staticConfig` and ended with `pictureControlSetPoolInitCount`.
Actually, parameters like `enableStrongIntraSmoothing`, `reduceCodingLoopCandidates`, `maxDpbSize` and other parametres were missing in the former version.

2) In the 2nd commit, `encoderBitDepth` parameter was assigned with the video size and the long expression `contextPtr->sequenceControlSetInstanceArray[instanceIndex]->sequenceControlSetPtr` was replaced by the local variable `sequenceControlSetPtr` .

PR #314, opened 2019-08-13
Fix infinite loop when encoding with codeVpsSpsPps enabled

Yeongjin-Jeong:
I have reported a bug that doesn't end encoding. (https://github.com/OpenVisualCloud/SVT-HEVC/issues/305)

This happens when using the EbH265EncStreamHeader API, after initializing the encoder with codeVpsSpsPps and before receiving the encoded packet.

In this case, some fields such as maxDpbSize, encoderBitDepth in SequenceControlSet that using by PacketizationKernel thread was not copied from main thread. So since the maxDbpSize value is 0 when creating VPS, WriteUvlc does infinite loop.

### DIFF
--- a/Source/Lib/Codec/EbEntropyCoding.c
+++ b/Source/Lib/Codec/EbEntropyCoding.c
@@ -4744,7 +4744,7 @@ static void WriteUvlc(
 	EB_U32 numberOfBits = 1;
 	EB_U32 tempBits = ++bits;
 
-	while (1 != tempBits)
+	while (tempBits > 1)
 	{
 		tempBits >>= 1;
 		numberOfBits += 2;


Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -4744,7 +4744,7 @@ static void WriteUvlc(
 	EB_U32 numberOfBits = 1;
 	EB_U32 tempBits = ++bits;
 
-	while (1 != tempBits)
+	while (1 != tempBits && tempBits > 0)
```
Austin-Hu: Looks good to me, but it's better to change to "while (tempBits > 1)". :)

Source/Lib/Codec/EbSequenceControlSet.c
```diff
@@ -239,6 +239,10 @@ EB_ERRORTYPE CopySequenceControlSet(
     dst->enableDenoiseFlag          = src->enableDenoiseFlag;                       writeCount += sizeof(EB_BOOL);
     dst->maxEncMode                 = src->maxEncMode;                              writeCount += sizeof(EB_U8);
 
+    dst->encoderBitDepth            = src->encoderBitDepth;                         writeCount += sizeof(EB_U32);
+    dst->rateControlMode            = src->rateControlMode;                         writeCount += sizeof(EB_U32);
+    dst->maxDpbSize                 = src->maxDpbSize;                              writeCount += sizeof(EB_U32);
```
Austin-Hu: These would conflict with PR #264 , so either needs to resolve the conflict after another one is merged.

Yeongjin-Jeong: Thanks to review :) This commit won't be needed after merging the PR https://github.com/OpenVisualCloud/SVT-HEVC/pull/264, so I'll drop it.

Source/Lib/Codec/EbEntropyCoding.c
Yeongjin-Jeong: Thanks for your suggestion. I applied it :)

tianjunwork:
@Yeongjin-Jeong Good fix, Thank you! Thanks @Austin-Hu for reviewing.

PR #320, opened 2019-08-16
Fix wrong bitstream indication for conformance to MainRext

Yeongjin-Jeong:
The encoderBitDepth field required to generate the H265 range extension profile indicator,
but it was not copied from the API. So for all bit depth encodings were indicated as 8bits in the 
range extension profile.

And the bitstream conforming to the format range extentions profiles should obey
the constraints defined by ITU-T H.265. But current code does not conform to 
the constraints in some cases. Such as, the previous generate a incorrect main-422-10 profile with 
the general_max_8bit_constraint_flag set to 1 when encoding YUV422/8bit with ConstrainedIntra mode.

The latest ITU-T H.265 revision referenced is (02/2018).

### DIFF
--- a/Source/Lib/Codec/EbEntropyCoding.c
+++ b/Source/Lib/Codec/EbEntropyCoding.c
@@ -4968,70 +4968,37 @@ static void CodeProfileTier(
         // "general_max_12bit_constraint_flag"
         WriteFlagCavlc(
            bitstreamPtr,
-           1);
+           (scsPtr->encoderBitDepth <= EB_12BIT));
 
         // "general_max_10bit_constraint_flag"
-        if(scsPtr->encoderBitDepth <= EB_10BIT || scsPtr->staticConfig.constrainedIntra == EB_TRUE)
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               1);
-        } else
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               0);
-        }
+        WriteFlagCavlc(
+           bitstreamPtr,
+           (scsPtr->encoderBitDepth <= EB_10BIT));
 
         // "general_max_8bit_constraint_flag"
-        //if(scsPtr->encoderBitDepth == EB_8BIT)
-        if(scsPtr->encoderBitDepth == EB_8BIT && (scsPtr->chromaFormatIdc == EB_YUV444 || scsPtr->staticConfig.constrainedIntra == EB_TRUE))
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               1);
-        } else
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               0);
-        }
+        WriteFlagCavlc(
+           bitstreamPtr,
+           (scsPtr->encoderBitDepth <= EB_8BIT && scsPtr->chromaFormatIdc != EB_YUV422));
 
         // "general_max_422chroma_constraint_flag"
-        if(scsPtr->chromaFormatIdc == EB_YUV422 || (scsPtr->chromaFormatIdc == EB_YUV420 && scsPtr->staticConfig.constrainedIntra == EB_TRUE))
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               1);
-        } else
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               0);
-        }
+        WriteFlagCavlc(
+           bitstreamPtr,
+           (scsPtr->chromaFormatIdc <= EB_YUV422));
 
         // "general_max_420chroma_constraint_flag"
-        if(scsPtr->chromaFormatIdc == EB_YUV420)
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               1);
-        } else
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               0);
-        }
+        WriteFlagCavlc(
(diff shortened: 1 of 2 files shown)


Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -4968,70 +4968,37 @@ static void CodeProfileTier(
...
-               1);
-        } else
-        {
-            WriteFlagCavlc(
-               bitstreamPtr,
-               0);
-        }
+        WriteFlagCavlc(
+           bitstreamPtr,
+           (scsPtr->encoderBitDepth <= EB_8BIT && scsPtr->chromaFormatIdc != EB_YUV422));
```
tianjunwork: Hi @Yeongjin-Jeong , what is the reason of adding "scsPtr->chromaFormatIdc != EB_YUV422"?

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -4968,70 +4968,37 @@ static void CodeProfileTier(
...
         WriteFlagCavlc(
            bitstreamPtr,
-           0);
+           (scsPtr->chromaFormatIdc <= EB_YUV400));
 
         // "general_intra_constraint_flag"
         WriteFlagCavlc(
            bitstreamPtr,
-           (scsPtr->staticConfig.constrainedIntra == EB_TRUE));
+           (scsPtr->intraPeriodLength == 0));
```
tianjunwork: I didn't find any related doc. Could you let me know how general_intra_constraint_flag is set? why set when gop is 1 instead of from staticConfig.constrainedIntra? Thank you.

Source/Lib/Codec/EbEntropyCoding.c
Yeongjin-Jeong: Hi @tianjunwork, Because there is no profile for 8-bit 4:2:2 in MainRext. So we should obey the constraints to main-422-10 profile.

Source/Lib/Codec/EbEntropyCoding.c
Yeongjin-Jeong: @tianjunwork I didn't find it too in ITU-T doc. But HM or x265 set this flag only when only using intra-picture prediction mode. 
(https://hevc.hhi.fraunhofer.de/HM-doc/_t_app_enc_cfg_8cpp_source.html : line 2253, https://github.com/videolan/x265/blob/master/source/encoder/level.cpp#L67)

tianjunwork: Thanks pointing out the reference code.

Source/Lib/Codec/EbEntropyCoding.c
tianjunwork: Thx. I find it in line 628 https://hevc.hhi.fraunhofer.de/HM-doc/_t_app_enc_cfg_8cpp_source.html

tianjunwork: But in TABLE I Profiles Defined in HEVC Version 1 and RExt and the Availability of Features of paper:
https://ieeexplore.ieee.org/document/7265015?arnumber=7265015
main 4:4:4 profile support 8bit and yuv 422.
Except HM, other implementations simple set scsPtr->encoderBitDepth <= EB_8BIT for general_max_8bit_constraint_flag

Yeongjin-Jeong: Yes, main-444 also covers 422/8 bits. But in x265 or Intel Meda SDK (when mapping their MainRext to vaapi's MainRext), they used main-422-10 instead of main-444.
(https://github.com/videolan/x265/blob/master/source/encoder/entropy.cpp#L399
https://github.com/Intel-Media-SDK/MediaSDK/blob/master/_studio/mfx_lib/encode_hw/h265/src/mfx_h265_encode_vaapi.cpp#L46)

tianjunwork:
Hi @lijing0010, could you also review this PR? Thank you.

Source/Lib/Codec/EbEntropyCoding.c
Austin-Hu: Hi @Yeongjin-Jeong ,

I think it's better to change with the condition (scsPtr->encoderBitDepth <= EB_8BIT) only to set general_max_8bit_constraint_flag, like what settings do for general_max_12bit_constraint_flag or general_max_10bit_constraint_flag, although the HEVC spec or the [table](https://ieeexplore.ieee.org/mediastore_new/IEEE/content/media/76/7372356/7265015/marpe.t1-2478707-large.gif) doesn't indicate to support 8-bit 4:2:2.

Source/Lib/Codec/EbEntropyCoding.c
Austin-Hu: According to "Table A.2 – Bitstream indications for conformance to format range extensions profiles" of HEVC spec, general_intra_constraint_flag should be set to 1 for intra (prediction) profiles or Still Picture ones. So this change is correct.

Source/Lib/Codec/EbEntropyCoding.c
Yeongjin-Jeong: Hi @Austin-Hu, Thanks for your suggestion.
If there is h265 parser that does not allow not standard constraints, It may be a problem. But most h265 parser(e.g. gstreamer) seem to match closet profile if bitstream contains not standard constraints. So I applied your review to this PR. :)

PR #323, opened 2019-08-19
Build.sh: Change to use bin sh instead of bash

1480c1:


tianjunwork:
Build error with clang, \["CMAKE\_EFLAGS=\\"-DBUILD\_SHARED\_LIBS=OFF\\""\]

1480c1:
Can you change the set -e to set-ex and rerun?

1480c1:
I can't seem to replicate, how did you run the script?

tianjunwork:
I rerun the job. Now it is passing.

PR #328, opened 2019-08-20
AMVP: injected the MV candidates with the correct ref list for AMVP.

Austin-Hu:


Austin-Hu:
This potential issue was found by @lijing0010. I just checked with one case (encMode = 1, because AMVP injection is only triggered with encMode <= 1), by making sure the playback looks good from the encoded bitstream, and no R2R issue. This change may need to be verified with more cases (encMode <= 1).

PR #336, opened 2019-08-22
New Test Cases

intelmark:
Added new test cases for MCTS, VBV, HDR, and Color
Format.

Modified the mcts test to adjust the maximum column
and row for the tile layout to be based on the frame
size (the number of luma pixels determines max rows and
max cols).

Also moved the actual determination of the max
column and row to where the video has already been
selected for the test.

Added -tile_slice-mode and -umv parameters for the
test.

Added decoding test for mcts to verify resulting bitstream
can be read.  This is similar to existing decode tests.

MCTS tests need to be run using updated HM decoder which
has a positive return code for when the decoding values.

### DIFF
--- a/Tests/SVT-HEVC_FunctionalTests.py
+++ b/Tests/SVT-HEVC_FunctionalTests.py
@@ -293,8 +293,8 @@ class EB_Test(object):
                         'EncoderColorFormat'                : '-color-format',
                         'TileRowCount'                      : '-tile_row_cnt',
                         'TileColCount'                      : '-tile_col_cnt',
-						'TileSliceMode'                     : '-tile_slice_mode',
-						'UnrestrictedMotionVector'          : '-umv',
+                        'TileSliceMode'                     : '-tile_slice_mode',
+                        'UnrestrictedMotionVector'          : '-umv',
                         'VbvMaxRate'                        : '-vbvMaxrate',
                         'VbvBufSize'                        : '-vbvBufsize',
                         'VbvBufInit'                        : '-vbvBufInit',
@@ -496,7 +496,6 @@ class EB_Test(object):
                         enc_params.update({cond: test_cond[cond]})
                         if not isinstance(test_cond[cond], list):
                             bitstream_name = bitstream_name + '_' + str(test_cond[cond])
-
                     # Check if sequence is supported with given combinations
                     error = self.check_seq_support(test_name, seq_name, enc_params)
                     if error != 0:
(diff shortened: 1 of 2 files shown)


Tests/SVT-HEVC_FunctionalTests.py
```diff
@@ -388,11 +404,11 @@ def check_seq_support(self, test_name, seq, enc_params):
             if encMode >= 12:
                 return -1
             if width*height < 3840*2160:
+                if encMode == 11:
+                    return -1
+            if width*height < 1920*1080:
                 if encMode == 10:
```
tianjunwork: encMode 11 is not allowed for  if width*height < 1920*1080:

Tests/SVT-HEVC_FunctionalTests.py
```diff
@@ -480,10 +496,12 @@ def run_test(self, test_name, test_params, enc_params, VBR, COMPARE):
                         enc_params.update({cond: test_cond[cond]})
                         if not isinstance(test_cond[cond], list):
                             bitstream_name = bitstream_name + '_' + str(test_cond[cond])
+
```
tianjunwork: remove empty lines

Tests/SVT-HEVC_FunctionalTests.py
```diff
@@ -496,14 +514,59 @@ def run_test(self, test_name, test_params, enc_params, VBR, COMPARE):
...
+                        # non-debug mode.. execute the encode command
                         exit_code = subprocess.call(enc_cmd, shell = True)
-                        if exit_code == 0 and test_name == 'decode_test':
-                            dec_cmd = enc_params['tools_dir'] + slash + dec_exe + " -b " + enc_params['bitstream_dir'] + slash + bitstream_name + '.265 > NUL'
-                            print(dec_cmd, file=open(test_name + '.txt', 'a'))
-                            exit_code = subprocess.call(dec_cmd, shell = True)
+                        if exit_code == 0:
+                            # For the decode and mcts tests see if the encoded file can be decoded.
+                            # HM decoder version should have a check to verify that motion vectors are
+                            # constrained to same tile, otherwise encoder fail.
```
tianjunwork: you mean decoder fail?

Tests/SVT-HEVC_FunctionalTests.py
```diff
@@ -770,6 +833,62 @@ def me_hme_test(self,seq_list):
...
+
+        mcts_rows = []
+        mcts_cols = []
+        for count in range(MCTS_ITER):
+            mcts_rows.append(0)
+            mcts_cols.append(0)
+        combination_test_params = { 'TileRowCount'     : mcts_rows,
+                                    'TileColCount'     : mcts_cols,
+									'UnrestrictedMotionVector' : [0],
+									'TileSliceMode' : [1]
```
tianjunwork: fix indent?

Tests/SVT-HEVC_FunctionalTests.py
intelmark: if width*height < 1920*1080 and encMode is set to 11, then it will caught in the check above:

if width*height< 3840*2160

tianjunwork: thank you:)

Issue #319, opened 2019-08-16
ffmpeg with libsvt_hevc fail to output video with specified resolution


andrewmabc:
I build ffmpeg with svt hevc using ffmpeg plugin patch. The following command outputs video with 960x544 instead of 960x540:

ffmpeg -c:v hevc -i h265_resolution_3840x2160_frame_rate_25_bitrate_34836000.mp4 -b:a 48000 -ar 44100 -g 250 -bufsize:v 1555200 -preset 9 -c:a libfdk_aac -c:v libsvt_hevc -maxrate:v 2073600 -profile:v 1 -vf "scale=w=960:h=540" -b:v 1036800 -movflags +faststart -strict -3 -vsync passthrough -f mp4 output.mp4

is it a bug?

Thanks!

tianjunwork:
Hi @jsunintel, could you help take a look? Thank you.

jsunintel:
540 is not a multiple of 8. SVT HEVC lib automatically changes it to 544. I don't think that's a bug.

tianjunwork:
540 is padded to 544. The new size and crop windows info is written to SPS.
![crop](https://user-images.githubusercontent.com/4551215/63289392-d9a4e200-c273-11e9-9962-c8a875210efd.PNG)
To understand stride:
https://docs.microsoft.com/en-us/windows/win32/medfound/image-stride

tianjunwork:
Padding is allowed for encoding. Not a SVT issue. Close.

jsunintel:
Hi Jing,
After solved last issue, we built trust with JD engineers. Today they reported another issues which he already submitted on github (https://github.com/OpenVisualCloud/SVT-HEVC/issues/319 ) and got response, but he didn’t satisfy. Do you have further explanation or work around solution to solve this kind of resolution issues? Thanks! 

Thanks and Best Regards
Xingyi Sun
Software Apps Engineer
IAGS CPDP CEE Cloud Software Engineering
Intel Corporation

lijing0010:
TianJun already explained it above.
From the spec:
"pic_width_in_luma_samples specifies the width of each decoded picture in units of luma samples.
pic_width_in_luma_samples shall not be equal to 0 and shall be an integer multiple of MinCbSizeY.
pic_height_in_luma_samples specifies the height of each decoded picture in units of luma samples.
pic_height_in_luma_samples shall not be equal to 0 and shall be an integer multiple of MinCbSizeY."

[...]

PR #338, opened 2019-08-23
Update documentation for -h -w

intelmark:
Adding This value should be a multiple of 8.

### DIFF
--- a/Docs/svt-hevc_encoder_user_guide.md
+++ b/Docs/svt-hevc_encoder_user_guide.md
@@ -19,7 +19,6 @@
 3. [ Sample Application Guide](#sample-application-guide)
     - 3.1 [Input Video Format](#input-video-format)
     - 3.2 [Compressed 10-bit format](#compressed-10-bit-format)
-    - 3.2 [Compressed 10-bit format](#compressed-10-bit-format)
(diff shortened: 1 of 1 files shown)


Docs/svt-hevc_encoder_user_guide.md
```diff
@@ -201,11 +201,11 @@ The resulting encoded bit stream file in binary format. If none specified, no ou
 
 >-w integer **[Required]**
 
-The width of each input image in units of picture luma pixels,  e.g. 1920
+The width of each input image in units of picture luma pixels,  e.g. 1920. This value should be a multiple of 8.
```
tianjunwork: No, it doesn't require to be multiple of 8.
refer to my answer:
https://github.com/OpenVisualCloud/SVT-HEVC/issues/319

intelmark:
Thanks for correcting.  Would this be more appropriate?:

+The width of each input image in units of picture luma pixels,  e.g. 1920. The frame width will have padding added so the final frame width is a multiple of 8.



+The height of each input image in units of picture luma pixels,  e.g. 1080. The frame height will have padding added so the final frame height is a multiple of 8.

tianjunwork:
+If width/length is not a multiple of 8, padding will be added.

PR #353, opened 2019-09-04
Don't set affinity if not necessary

lijing0010:
For non-numa architecture, users can use taskset to specify cores

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -1020,8 +1020,10 @@ void EbSetThreadManagementParameters(
     if (numGroups == 1) {
         EB_U32 lps = configPtr->logicalProcessors == 0 ? numLogicProcessors:
             configPtr->logicalProcessors < numLogicProcessors ? configPtr->logicalProcessors : numLogicProcessors;
-        for(EB_U32 i=0; i<lps; i++)
-            CPU_SET(lpGroup[0].group[i], &groupAffinity);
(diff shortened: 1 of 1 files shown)


Source/Lib/Codec/EbEncHandle.c
```diff
@@ -1020,8 +1020,10 @@ void EbSetThreadManagementParameters(
     if (numGroups == 1) {
         EB_U32 lps = configPtr->logicalProcessors == 0 ? numLogicProcessors:
             configPtr->logicalProcessors < numLogicProcessors ? configPtr->logicalProcessors : numLogicProcessors;
-        for(EB_U32 i=0; i<lps; i++)
-            CPU_SET(lpGroup[0].group[i], &groupAffinity);
+            if (configPtr->targetSocket != -1) {
```
tianjunwork: numGroups == 1 means there is only 1 socket.
Affinity setting here only takes -lp value(configPtr->logicalProcessors) into consideration.
configPtr->targetSocket is any way ignored here when calculating lps in line 1021.
To use taskset, no -lp setting should be config.

tianjunwork: Might revert this PR?
Otherwise, to set affinity for single socket CPU, -ss N is needed to have the -lp N logic running, which doesn't make sense.

lijing0010: I think may need some more codes here.
without this patch, we can't set the affinity with taskset from command line for single socket CPU

PR #92, opened 2019-02-04
Add AppVeyor CI/CD

EwoutH:
This PR adds continuous integration & delivery with AppVeyor. SVT-HEVC gets build with Microsoft Visual Studio 2017 and Yasm 1.3.0. AppVeyor publishes `SvtHevcEnc.dll`, `SvtHevcEnc.exp`, `SvtHevcEnc.lib`, `SvtHevcEncApp.exe` and `SvtHevcEncSimpleApp.exe` in the Artifacts tab ([example](https://ci.appveyor.com/project/EwoutH/svt-hevc/builds/22122388/artifacts)).

AppVeyor needs to be enabled on the [GitHub Marketplace](https://github.com/marketplace/appveyor) or on [ci.AppVeyor.com](https://ci.appveyor.com/project/OpenVisualCloud/SVT-HEVC).

When enabled, AppVeyor publishes the artifacts with each build. When tagging a new (pre)release, AppVeyor will also upload these files to GitHub, creating a release page [like this](https://github.com/EwoutH/SVT-HEVC/releases/tag/v1.2.1-test1).

To enable uploading to GitHub Releases, a maintainer needs to update this patch with his personal authentication token. You can generate one on https://github.com/settings/tokens/new (select `public_repo` as scope), and then encrypt it with https://ci.appveyor.com/tools/encrypt. Replace the current token on line 21 (after `secure:`) in the `appveyor.yml` file.

PR #361, opened 2019-09-12
Update AppVeyor to only deploy VS19 Release build to GitHub Releases

EwoutH:
Instead of trying to push all 5 builds to GitHub on a release, it now only deploys the Release build from Visual Studio 2019.

Aside from that, it seems that the `secure auth_token` is not working, since [AppVeyor did fail](https://ci.appveyor.com/project/OpenVisualCloud/svt-hevc/builds/27345109) on the v1.4.1 tag.
```
Deploying using GitHub provider
Creating "v1.4.1" release for repository "OpenVisualCloud/SVT-HEVC" tag "v1.4.1" commit "02fd1261966acfae6b363d8213710ef7505f0f31"...Error creating GitHub release: Provider setting not found or it's value is empty. If secure setting is used please check that value was encrypted (or YAML was exported) while being logged under correct account.
Parameter name: auth_token
```
I think it wasn't changed (correctly) when #92 got merged. @hassount or @tianjunwork, could you check if the token is correct (line 116 in the `appveyor.yml`) and if not change it according to the instructions in #92?

Thanks @michaellarabel for noticing this issue in https://github.com/phoronix-test-suite/test-profiles/pull/98#issuecomment-530414442.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -113,7 +113,7 @@ deploy:
   - provider: GitHub
     artifact: $(APPVEYOR_PROJECT_NAME)
     auth_token:
-      secure: "hY5Mk6KOwgQ97TzEBsM7Woqr1ZIm5QTvHg8EvxMV1x8j3wk/3mNBMqWFFbEIBK0i"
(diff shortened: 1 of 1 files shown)


tianjunwork:
@EwoutH sorry for the late response. I am in vacation right now. Thanks @intelmark and @1480c1 for helping out.
Could you add the sign-off line in the commit msg? Thank you.

PR #362, opened 2019-09-12
Fixed the encoding crashing issue for some legacy CPUs.

Austin-Hu:
By checking whether the processor supports XGETBV instruction or not,
beforeing invoking it to check ZMM/YMM/XMM states are enabled in XCR0
register.

### DIFF
--- a/Source/Lib/ASM_AVX2/EbIntraPrediction_Intrinsic_AVX2.c
+++ b/Source/Lib/ASM_AVX2/EbIntraPrediction_Intrinsic_AVX2.c
@@ -1873,115 +1873,31 @@ void IntraModeVerticalChroma_AVX2_INTRIN(
 
     // Jing: 
     // TODO: add size == 32 for 444
-    if (!skip) {
-        if (size == 32) {
-            __m256i xmm0;
-            EB_U64 size_to_write;
-            EB_U32 count;
+    if (!skip && (size == 32)) {
+        __m256i xmm0;
+        EB_U64 size_to_write;
+        EB_U32 count;
 
-            // Each storeu calls stores 32 bytes. Hence each iteration stores 8 * 32 bytes.
-            // Depending on skip, we need 4 or 2 iterations to store 32x32 bytes.
-            size_to_write = 4 >> (skip ? 1 : 0);
-            pStride = pStride << (skip ? 1 : 0);
+        // Each storeu calls stores 32 bytes. Hence each iteration stores 8 * 32 bytes.
+        // Depending on skip, we need 4 or 2 iterations to store 32x32 bytes.
+        size_to_write = 4 >> (skip ? 1 : 0);
+        pStride = pStride << (skip ? 1 : 0);
 
-            xmm0 = _mm256_loadu_si256((__m256i *)(refSamples + topOffset));
+        xmm0 = _mm256_loadu_si256((__m256i *)(refSamples + topOffset));
 
-            for (count = 0; count < size_to_write; count ++) {
-                _mm256_storeu_si256((__m256i *)(predictionPtr), xmm0);                    
-                _mm256_storeu_si256((__m256i *)(predictionPtr + pStride), xmm0);          
-                _mm256_storeu_si256((__m256i *)(predictionPtr + 2 * pStride), xmm0);      
-                _mm256_storeu_si256((__m256i *)(predictionPtr + 3 * pStride), xmm0);         
+        for (count = 0; count < size_to_write; count ++) {
+            _mm256_storeu_si256((__m256i *)(predictionPtr), xmm0);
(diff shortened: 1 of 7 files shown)


1480c1:
> Because we don't have such CPU to debug in hand, or we'd try to find one in some day. And could you help debug this issue on your X3430 platform, if you have more bandwidth, @1480c1 ? For example, try to load refSamples (without topOffset added) with _mm_loadu_si128, or try to load with other intrinsic (such as _mm_loadu_si64) to see any difference. Thanks!

Sorry for not seeing this, since it's an edit, it didn't notify anything. 

I can try, however, I'm not proficient in understanding simd instructions atm. 

Would it be preferable to have a way for you to ssh into it?

Austin-Hu:
> > Because we don't have such CPU to debug in hand, or we'd try to find one in some day. And could you help debug this issue on your X3430 platform, if you have more bandwidth, @1480c1 ? For example, try to load refSamples (without topOffset added) with _mm_loadu_si128, or try to load with other intrinsic (such as _mm_loadu_si64) to see any difference. Thanks!
> 
> Sorry for not seeing this, since it's an edit, it didn't notify anything.
> 
> I can try, however, I'm not proficient in understanding simd instructions atm.
> 
> Would it be preferable to have a way for you to ssh into it?

@1480c1 , sorry, I should have added my comments by appending a new one, rather than editting.

I'm not proficient but I'd like to try to after completing my current tasks. Could you share me your SSH info, and are you using Linux or Windows? Thanks!

1480c1:
I can setup an ssh account on the machine, it's running Ubuntu 19.04. currently about to have an exam, so will not be able to do it immediately

Austin-Hu:
> I can setup an ssh account on the machine, it's running Ubuntu 19.04. currently about to have an exam, so will not be able to do it immediately

OK, let me try to log in it.

1480c1:
I sent you an email to the one in the commit with the information

Austin-Hu:
Hi @1480c1 ,

As for the latter "_mm_loadu_si128" crashing issue, I think its root cause is that, in the implementation of some (or most) kernel threads, using AVX and SSE intrinsics were mixed together, such as function IntraModeVerticalChroma_AVX2_INTRIN(). So with your X3430 CPU without AVX intrinsic, the IntraVerticalChroma_funcPtrArray function pointer array would point to the IntraModeVerticalChroma_AVX2_INTRIN() code with PREAVX2_MASK, and would go through the section using AVX intrinsic (the 256bit version, or the 128bit one compiled with -mavx2 gcc option) which it shouldn't have gone.

The way to address this issue, is to separate the SSE/AVX intrinsic usings into different files, compiled with corresponding gcc x86 [options](http://gcc.gnu.org/onlinedocs/gcc/x86-Options.html). And the most ideal way I was considering is to use the "CPU dispatch" mechanism mentioned in Section 13.5 of ["Optimizing Software in C++"](https://www.agner.org/optimize/optimizing_cpp.pdf). But it would be feasible for the SVT encoder project, because the function pointer of most kernel threads changes during runtime. For example, using either SSE or AVX in IntraModeVerticalChroma_AVX2_INTRIN() depends on the Chroma PU size, rather than being set during encoder initialization stage.

So I submitted an experimental [commit](https://github.com/OpenVisualCloud/SVT-HEVC/pull/362/commits/d708ea7066684ee17d9155d312e90b6d695a9b5f), to prove that the file using SSE and being compiled with -mavx2 would cause the crashing issue in your CPU withough AVX supported. And with the commit applied, the "Illegal instruction" issue wouldn't happen in IntraModeVerticalChroma_AVX2_INTRIN(), but it occurred again in another following function IntraModeDCChroma_AVX2_INTRIN().

In order to fully resolve this issue, we need too much similar seperation (physical effort actually) for all the implementations of kernel threads which mixed different intrinsics, not only for IntraModeVerticalChroma_AVX2_INTRIN(). So let's set it as low priority and refine those codes in the future to support legacy CPUs.

Thanks!

Austin-Hu:
BTW, I suggest to review and merge the 1st [commit](https://github.com/OpenVisualCloud/SVT-HEVC/pull/362/commits/ae823f2f81fab5399b058be61fbe9b8a711904d7) for both SVT-HEVC and SVT-VP9, as it's a reasonable fix. @1480c1 @tianjunwork @lijing0010

Austin-Hu:
To merge the PR #376 which is a correct fix with single commit for this issue.

As for the experimental [commit](https://github.com/OpenVisualCloud/SVT-HEVC/pull/362/commits/d708ea7066684ee17d9155d312e90b6d695a9b5f), which can be referenced instead [here](https://github.com/Austin-Hu/SVT-HEVC/commit/d708ea7066684ee17d9155d312e90b6d695a9b5f), set it as low priority.

Issue #373, opened 2019-09-27
problem when reading from pipe


calvinckfong:
Good day!  When I follow the instruction in readme to read from pipe, there is a problem in v1.4.0 or above.  I used the follow command and several bytes shifted are observed.  There is no such problem in v1.3.0.  Is there any extra option I needed?  Thanks!
 
`$FFMPEG -hide_banner -loglevel 0 -i $SRC -pix_fmt yuv420p - | $SVTHEVC -i stdin -w $WIDTH -h $HEIGHT -n $FRAME_NUM -b $OUT.hevc`

![svt-hevc-500](https://user-images.githubusercontent.com/16603747/65736816-f2c06e80-e10e-11e9-839a-7c3c1e2c96e9.png)

Selur:
Also y4m pipe doesn't work,
[code]ffmpeg -y -loglevel fatal -noautorotate -nostdin -threads 8 -i "F:\TestClips&Co\Test-AAC-5.1.mp4" -map 0:0 -an -sn -vf zscale=rangein=tv:range=tv -pix_fmt yuv420p10le -strict -1 -vsync 0  -f yuv4mpegpipe - | SvtHevcEncApp -encMode 8 -i stdin -fps-num 25 -fps-denom 1 -n 834 -profile 2  -b "E:\Temp\Test-AAC-5.1_new_19_36_40_7610_01.265"[/code]
fails until "-w ... -h ..." is added, so in case this should work: it doesn't.
(used: commit 3d68dd8397400bd70b48710f1b3b43a0bb3872f3)

Only way to get proper output files is atm. is not to use pipes.

intelmark:
@calvinckfong  Could you provide a command line without the variables so we can see the exact parameters ($WIDTH, $HEIGHT, $SRC etc.) as well as the input stream?
thanks..

intelmark:
@selur - Could you open a new issue on this problem (unless you strong believe it to be related the first issue)?

Selur:
Since I strongly believe that in general pipe input is broken and my problem is related to the first issue:

```
[...]

PR #380, opened 2019-10-07
Fixed the Y4M header analysis issue in SvtHevcEncApp.

Austin-Hu:
The issue was caused by the definition of EB_STRNCPY invoking strncpy_ss
with sizeof(dst). sizeof() an array can return the memory size of the
array, while sizeof a pointer pointing to an array would return the size
of the pointer (8 in 64bit machine, and 4 in 32bit one). So that dmax of
strncpy_ss was always 8 or 4, and would cause the byte assignment issue
once the size of source was greater or equal than dmax.


Austin-Hu:
Fixed Issue #373 .

PR #381, opened 2019-10-08
Fix 8x8 CU scan index error in function `DeriveLcuScore()`.

luofalei:
The `cu8x8Index` here is wrong for `lcuParams->rasterScanCuValidity[cu8x8Index]` check.


tianjunwork:
Hi @luofalei Failei, Thank you for submitting the patch. Besides reviewing the code change, how can we verify the difference of bitstream? Could you share the example of command line which triggers the issue?

luofalei:
@tianjunwork Currently I haven't produced any different results. This commit seems hard to triger different bit-streams since it only influenced the score of LCUs across the picture boudaries, while some CU skip algorithms were off for CUs in these LCUs.

tianjunwork:
@luofalei thank you, no problem:)

PR #382, opened 2019-10-08
Fix LumaIntraMode storage error in function `LinkBdptoMd()` and `LinkMdtoBdp()`

luofalei:
In function `LinkBdptoMd()` and `LinkMdtoBdp()`, when one CU is coded in `Intra2Nx2N` mode, the `predictionModeFlag` was passed as `intraLumaMode` to update the neighboring information arrays.

This commit fixed the above error.

### DIFF
--- a/Source/Lib/Codec/EbProductCodingLoop.c
+++ b/Source/Lib/Codec/EbProductCodingLoop.c
@@ -3806,7 +3806,7 @@ EB_EXTERN EB_ERRORTYPE LinkBdptoMd(
     LargestCodingUnit_t     *lcuPtr,
     ModeDecisionContext_t   *contextPtr)
 {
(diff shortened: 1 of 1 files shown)


tianjunwork:
Hi @luofalei Failei, Thank you for submitting the patch. Besides reviewing the code change, how can we verify the difference of bitstream? Could you share the example of command line which triggers the issue?

luofalei:
@tianjunwork Hi, the following command can produce different bitstream.
```
../../Bin/Release/SvtHevcEncApp -i BasketballPass_416x240_50.yuv -n 500 -w 416 -h 240 -fps 50 -intra-period 49 -q 32 -bit-depth 8 -encMode 4 -b test.hevc
```

PR #341, opened 2019-08-26
FFmpeg Plugin: Add setup steps for Windows

jsunintel:

PR #386, opened 2019-10-09
FFmpeg doc

intelmark:
Added instruction for running the second patch file.

### DIFF
--- a/ffmpeg_plugin/README.txt
+++ b/ffmpeg_plugin/README.txt
@@ -50,6 +50,7 @@ For Both:
 - cd ffmpeg
 - git checkout release/4.2
 - git am ../SVT-HEVC/ffmpeg_plugin/0001*.patch
+- git am ../SVT-HEVC/ffmpeg_plugin/0002*.patch
 - ./configure --enable-libsvthevc
 - make -j $(nproc)
 


ffmpeg_plugin/README.txt
```diff
@@ -50,6 +50,7 @@ For Both:
 - cd ffmpeg
 - git checkout release/4.2
 - git am ../SVT-HEVC/ffmpeg_plugin/0001*.patch
+- git am ../SVT-HEVC/ffmpeg_plugin/0002*.patch
```
tianjunwork: Does 0002* apply to 4.2? I remember there are conflicts?

intelmark: The patch seems to be correctly applied in the 4.2 branch and help information is correctly displayed from ffmpeg.

I noticed that the tune description should be updated:
+@item tune
+Set quality tuning mode. Can assume one of the following possible values:
+
+@table @samp
+@item sq
+Visually optimized mode
+@item oq
+PSNR / SSIM optimized mode
+@item vmaf
+VMAF optimized mode
+@end table

tianjunwork: Then it is the master that can't apply? In #341 0002 is removed. Do you remember the reason?

intelmark: Yes, tt was changed in 341:

Author:			SUN, Jing <jing.a.sun@intel.com>
Date:			1 month ago (8/28/2019 12:41:08 PM)
Committer:		Mark Feldman <mark.feldman@intel.com>
Commit hash:	da3f4967a8ecf6535a168e08bcb71410b978d14d
Parent:			694080fffb

from:  git apply ../SVT-HEVC/ffmpeg_plugin/*.patch
to: git am ../SVT-HEVC/ffmpeg_plugin/0001*.patch

not sure why...

jsunintel:
Please do not apple the second patch. It's a patch to Docs, which is not necessary. The master and the current release branches tend to diff and the patch can only be made based on one. Regularly applying this patch makes users confused.

tianjunwork:
Thanks Jing for your comment.

intelmark:
Thanks Jing as well, but is there any other mechanism to update the documentation for the plugin.  

(e.g.) How does the user know that tune is depreciated, etc.

PR #397, opened 2019-10-16
10bit compressed check

intelmark:
Adding a configuration check that the compressed 10 bit format can only
be used if the encoder bit depth is also 10 bits.

### DIFF
--- a/Source/App/EbAppConfig.c
+++ b/Source/App/EbAppConfig.c
@@ -955,6 +955,12 @@ static EB_ERRORTYPE VerifySettings(EbConfig_t *config, uint32_t channelNumber)
         return_error = EB_ErrorBadParameter;
     }
 
+    if (config->encoderBitDepth != 10 && config->compressedTenBitFormat == EB_TRUE)
+    {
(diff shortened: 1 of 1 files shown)


Source/App/EbAppConfig.c
```diff
@@ -955,6 +955,12 @@ static EB_ERRORTYPE VerifySettings(EbConfig_t *config, uint32_t channelNumber)
         return_error = EB_ErrorBadParameter;
     }
 
+    if (config->encoderBitDepth != 10 && config->compressedTenBitFormat == EB_TRUE)
+    {
+        fprintf(config->errorLogFile, "SVT [Error]: Instance %u: Compressed 10 bit format inconsistent with encoder bit depth\n", channelNumber + 1);
```
tianjunwork: I always think "encoder bit depth" is no better than "encoderBitDepth" nor "-bit-depth". Same for "Compressed 10 bit format".
What's is your thought?

intelmark: If you are asking about how we format the variable name or describe the feature, I'm OK with either.  Is there a specific change you think we should make?

PR #399, opened 2019-10-25
Enabling segment override for each LCU separatelly

tszumski:
Hi, I want to introduce new feature for HEVC
The point is to give users ability to change QP and densilty for each LCU and image separately.
This feature is rather for very specific use-case like surveillance systems, where objects like pedestrians or vehicles are detected and tracked (encoded with higher quality) while background is encoded with worse quality. Thanks to encoding only key areas with good quality bitstream is smaller and still have high value

I added new parameter "-segment-ov-file" which read qp values from file and send it to encoder (as example)

I am open to any suggestions.

### DIFF
--- a/Config/SVTSegmentOvFile.txt
+++ b/Config/SVTSegmentOvFile.txt
@@ -1,0 +1,19 @@
+# Copyright(c) 2019 Intel Corporation 
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#====================== Segment Override File ======================
+# One input frame have to be preceded by the number of regions
+# Each region consist of 6 numbers (X1,Y1,X2,Y2, data, mode) i.e (X1,Y1)top left and (X2,Y2)bottom right corner
+# Mode parameter have to be in range [1..5]. This allow user to specify what to override by the [data]
+#    1.   Set qp value for each region, [data] is clipped to [0:51]
+#    2.   Change qp value i.e increase or decrease by the [data] clipped to [-25:25]
+#    3.   Use this mode to modify not only quality but also blocks splitting, behaviour depends on parameter 1 and 2
+#    4.   Use this mode to modify blocks splitting i.e increase or decrease by the [data] clipped to [-25:25], negative value increase block division
+#    5.   Update first stage of quantization in range [-7;7] positive value reduce bitrate, negative increase
+3
+0 0 500 500 50 1
+500 500 1000 0 -20 4
+500 500 1000 1000 20 4
+2
+0 0 500 500 20 1
+0 0 300 200 20 1
--- a/Source/API/EbApi.h
+++ b/Source/API/EbApi.h
@@ -38,6 +38,33 @@ extern "C" {
 #define EB_NON_REF_PICTURE   4
 #define EB_INVALID_PICTURE   0xFF
 
+#define EB_SEGMENT_BLOCK_SIZE 64
+    typedef struct
+    {
+        uint8_t ovFlags;
+        int8_t  qpOv;
+        int8_t  deblockOv;
+        int8_t  filterOv;
+    } SegmentOverride_t;
+
+    typedef enum EB_OV_FLAGS
+    {
+        /* Do no ovveride default values */
+        EB_NO_OV              = 0,
+        /* Set qp value in range [0:51] */
+        EB_QP_OV_DIRECT       = 1 << 0,
+        /* Change original qp value in range [-25:25] */
+        EB_QP_OV_DELTA        = 1 << 1,
+        /* Use qpOv for deblock, behaviour depends on EB_QP_OV* flag */
+        EB_DENSITY_QP_OV      = 1 << 2,
+        /* Use deblockOv for deblock in range [-25:25];
+        * positive value decrease density, negative increase */
+        EB_DENSITY_DEBLOCK_OV = 1 << 3,
+        /* Change first stage of quantization in range [-7;7];
+        * positive value reduce bitrate, negative increase */
+        EB_TU_FILTER_OV       = 1 << 4,
+    } EB_OV_FLAGS;
+
     typedef struct EB_BUFFERHEADERTYPE
     {
         // EB_BUFFERHEADERTYPE size
@@ -72,6 +99,7 @@ extern "C" {
         uint32_t naluPayloadType;
         uint8_t* naluBase64Encode;
 
+        SegmentOverride_t *segmentOvPtr;
     } EB_BUFFERHEADERTYPE;
 
     typedef struct EB_COMPONENTTYPE
@@ -629,6 +657,10 @@ typedef struct EB_H265_ENC_CONFIGURATION
     uint16_t                whitePointX, whitePointY;
     uint32_t                maxDisplayMasteringLuminance;
     uint32_t                minDisplayMasteringLuminance;
+
+    /* Flag to enable SegmentOverride overwrite per LCU
+    * Default is 0. */
+    uint32_t                segmentOvEnabled;
 
 } EB_H265_ENC_CONFIGURATION;
 
--- a/Source/App/EbAppContext.c
+++ b/Source/App/EbAppContext.c
@@ -273,6 +273,8 @@ EB_ERRORTYPE CopyConfigurationParameters(
     callbackData->ebEncParameters.maxDisplayMasteringLuminance = config->maxDisplayMasteringLuminance;
     callbackData->ebEncParameters.minDisplayMasteringLuminance = config->minDisplayMasteringLuminance;
 
+    callbackData->ebEncParameters.segmentOvEnabled = config->segmentOvEnabled;
+
     return return_error;
 
 }
@@ -379,6 +381,13 @@ EB_ERRORTYPE AllocateInputBuffers(
         // Assign the variables
         callbackData->inputBufferPool->pAppPrivate = NULL;
         callbackData->inputBufferPool->sliceType   = EB_INVALID_PICTURE;
+    }
+
+    if (callbackData->ebEncParameters.segmentOvEnabled) {
+        size_t pictureWidthInLcu = (config->sourceWidth + EB_SEGMENT_BLOCK_SIZE - 1) / EB_SEGMENT_BLOCK_SIZE;
+        size_t pictureHeightInLcu = (config->sourceHeight + EB_SEGMENT_BLOCK_SIZE - 1) / EB_SEGMENT_BLOCK_SIZE;
+        size_t lcuTotalCount = pictureWidthInLcu * pictureHeightInLcu;
+        EB_APP_MALLOC(SegmentOverride_t*, callbackData->inputBufferPool->segmentOvPtr, sizeof(SegmentOverride_t) * lcuTotalCount, EB_N_PTR, EB_ErrorInsufficientResources);
     }
 
     return return_error;
--- a/Source/Lib/Codec/EbCodingLoop.c
+++ b/Source/Lib/Codec/EbCodingLoop.c
@@ -2272,6 +2272,13 @@ void SetPmEncDecMode(
 
         }
     }
+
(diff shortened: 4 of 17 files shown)


Config/SVTSegmentOvFile.txt
```diff
@@ -0,0 +1,25 @@
+# Copyright(c) 2019 Intel Corporation 
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#====================== Segment Override File ======================
+# If the number of frames * blockPerFrame > number of QPs, the application will loop to the beginning of the file
+# Diffrence between largest and smallest value should be less or equal than 25
+# Below values are valid for 1080p video (30x17)
```
Austin-Hu: Hi @tszumski ,

Users can set the values for the specific resolution of input frame, but we'd better consider dynamical setting, for some scenarios such as resolution change. So how about:
    1. define an array with the values: [27, 29, 31, 33, 35, 37, 39, 41, 43, 45, 47, 49, 51];
    2. Repeat assgign each element of the array into another dynamic one, with the times = picWidth / LCU_SIZE;
    3. Loop and repeat assign the next element of the array into the dynamic one. (loopCount = picHeight / LCU_SIZE).

Maybe in this way, we don't need to load from file? Unless that users would like to set other customized values.

Source/API/EbApi.h
```diff
@@ -38,6 +38,33 @@ extern "C" {
 #define EB_NON_REF_PICTURE   4
 #define EB_INVALID_PICTURE   0xFF
 
+#define EB_SEGMENT_BLOCK_SIZE 64
+    typedef struct
+    {
+        uint8_t ovFlags;
+        int8_t  qpOv;
+        int8_t  deblockOv;
+        int8_t  filterOv;
```
Austin-Hu: I don't see where deblockOv and filterOv are initialized or assigned. Should be in the Loop Filter process?

Source/API/EbApi.h
```diff
@@ -38,6 +38,33 @@ extern "C" {
...
+        /* Change original qp value in range [-26:25] */
+        EB_QP_OV_DELTA        = 1 << 1,
+        /* Use qpOv for deblock, behaviour depends on EB_QP_OV* flag */
+        EB_DENSITY_QP_OV      = 1 << 2,
+        /* Use deblockOv for deblock in range [-25:25];
+        * positive value decrease density, negative increase */
+        EB_DENSITY_DEBLOCK_OV = 1 << 3,
+        /* Change first stage of quantization in range [-7;7];
+        * positive value reduce bitrate, negative increase */
+        EB_TU_FILTER_OV       = 1 << 4,
```
Austin-Hu: Setting the 2 ones above for the bitmap hasn't been completed yet?

Source/Lib/Codec/EbCodingLoop.c
```diff
@@ -2272,6 +2272,13 @@ void SetPmEncDecMode(
 
         }
     }
+
+    if (pictureControlSetPtr->ParentPcsPtr->segmentOvArray != NULL && sequenceControlSetPtr->staticConfig.segmentOvEnabled) {
+		SegmentOverride_t* segmentOvPtr = pictureControlSetPtr->ParentPcsPtr->segmentOvArray;
```
Austin-Hu: Please use 4 spaces instead of tabs for indents, and in other changes as well.

Source/Lib/Codec/EbCodingLoop.c
```diff
@@ -2272,6 +2272,13 @@ void SetPmEncDecMode(
...
+        if(segmentOvPtr[lcuIndex].ovFlags & EB_TU_FILTER_OV)
```
Austin-Hu: A space should be added after "if".

Source/Lib/Codec/EbCodingLoop.c
```diff
@@ -2272,6 +2272,13 @@ void SetPmEncDecMode(
...
+            contextPtr->pmpMaskingLevelEncDec= CLIP3(0, 7, contextPtr->pmpMaskingLevelEncDec + segmentOvPtr[lcuIndex].filterOv);
```
Austin-Hu: A space should be added before "=". And similiar coding style flaws in other changes.

Source/Lib/Codec/EbCodingLoop.c
```diff
@@ -3145,11 +3159,11 @@ EB_EXTERN void EncodePass(
     }
 
 
-    EB_BOOL useDeltaQp = (EB_BOOL)(sequenceControlSetPtr->staticConfig.improveSharpness || sequenceControlSetPtr->staticConfig.bitRateReduction);
+    EB_BOOL useDeltaQp = (EB_BOOL)(sequenceControlSetPtr->staticConfig.improveSharpness || sequenceControlSetPtr->staticConfig.bitRateReduction || sequenceControlSetPtr->staticConfig.segmentOvEnabled);
```
Austin-Hu: Is it better to check (pictureControlSetPtr->ParentPcsPtr->segmentOvArray[lcuPtr->index].ovFlags & EB_QP_OV_DELTA)?

Source/Lib/Codec/EbCodingLoop.c
```diff
@@ -3145,11 +3159,11 @@ EB_EXTERN void EncodePass(
...
 
     EB_BOOL singleSegment = (sequenceControlSetPtr->encDecSegmentColCountArray[pictureControlSetPtr->temporalLayerIndex] == 1) && (sequenceControlSetPtr->encDecSegmentRowCountArray[pictureControlSetPtr->temporalLayerIndex] == 1);
 
-    EB_BOOL useDeltaQpSegments = singleSegment ? 0 : (EB_BOOL)(sequenceControlSetPtr->staticConfig.improveSharpness || sequenceControlSetPtr->staticConfig.bitRateReduction);
+    EB_BOOL useDeltaQpSegments = singleSegment ? 0 : (EB_BOOL)(sequenceControlSetPtr->staticConfig.improveSharpness || sequenceControlSetPtr->staticConfig.bitRateReduction || sequenceControlSetPtr->staticConfig.segmentOvEnabled);
```
Austin-Hu: The same with above.

Source/App/EbAppContext.c
```diff
@@ -381,6 +383,13 @@ EB_ERRORTYPE AllocateInputBuffers(
         callbackData->inputBufferPool->sliceType   = EB_INVALID_PICTURE;
     }
 
+    if (callbackData->ebEncParameters.segmentOvEnabled) {
+        size_t pictureWidthInLcu=(config->sourceWidth + EB_SEGMENT_BLOCK_SIZE - 1) / EB_SEGMENT_BLOCK_SIZE;
+        size_t pictureHeightInLcu=(config->sourceHeight + EB_SEGMENT_BLOCK_SIZE - 1) / EB_SEGMENT_BLOCK_SIZE;
+        size_t lcuTotalCount = pictureWidthInLcu * pictureHeightInLcu;
+        EB_APP_MALLOC(SegmentOverride_t*, callbackData->inputBufferPool->segmentOvPtr, sizeof(SegmentOverride_t) * lcuTotalCount, EB_N_PTR, EB_ErrorInsufficientResources);
```
Austin-Hu: Need to free them as well, in the DeInitEncoder() procedure.

tszumski: As far as i understand the code and usage of EB_APP_MALLOC/EB_MALLOC, if I will use those macro,  allocator will free all allocated memory automatically in DeInitEncoder(). 

Am I right?

Source/Lib/Codec/EbCodingLoop.c
tszumski: pictureControlSetPtr->ParentPcsPtr->segmentOvArray could be NULL (memory is allocated only if sequenceControlSetPtr->staticConfig.segmentOvEnabled is True ).  To make code as simple as possible I would leave this kind of change as is

Source/Lib/Codec/EbCodingLoop.c
tszumski: pictureControlSetPtr->ParentPcsPtr->segmentOvArray could be NULL (memory is allocated only if sequenceControlSetPtr->staticConfig.segmentOvEnabled is True ). To make code as simple as possible I would leave this kind of change as is

Config/SVTSegmentOvFile.txt
tszumski: The point is to allow user to set customized values for each frame and LCU separately.
I have another idea to read values in different way:
    1. One input frame data could be preceded by the number of regions
    2. Each region will consist of 6 values (X1,Y1,X2,Y2, value, mode) i.e top left and bottom right corners 

As example:
"""
2 //for frame with idx=0, set 2 separate regions
0 0 300 300 40 1 // set for region (from point 0,0 to 300,300) value 40 in mode 1 (i.e EB_QP_OV_DIRECT )
900 600 1200 900 20 2// set for region (from point 900,600 to 1200,900) value 20 in mode 2 (i.e EB_QP_OV_DELTA )
1 //for frame with idx=1 set 1 region
0 0 300 300 30 1 // set for region (from point 0,0 to 300,300) value 30 in mode 1 (i.e EB_QP_OV_DIRECT )

If we reached the end of file, loop over again


How about this idea?

tszumski:
Hi @Austin-Hu 

Thanks for comments, and please check my answers.

Source/App/EbAppContext.c
Austin-Hu: I found it that EbDeinitEncoder() will free all the recorded memory resources.

Source/Lib/Codec/EbCodingLoop.c
Austin-Hu: So how about
```
EB_BOOL useDeltaQp = (EB_BOOL)(sequenceControlSetPtr->staticConfig.improveSharpness || sequenceControlSetPtr->staticConfig.bitRateReduction ||
    (sequenceControlSetPtr->staticConfig.segmentOvEnabled && (pictureControlSetPtr->ParentPcsPtr->segmentOvArray[lcuPtr->index].ovFlags & EB_QP_OV_DELTA)));
```
?

Because delta QP is subset of your seg OV feature, and this checking would make it more accurate.

Source/Lib/Codec/EbCodingLoop.c
Austin-Hu: Same with above.

Config/SVTSegmentOvFile.txt
Austin-Hu: I'm unclear about the typical use case or scenario. Do the users need to prepare such file with QP specificed for each LCU or region of each frame? Or is there any way (tool or module) to be used to inference different QP values for the LCUs/regions of input frames during encoding?

My suggestion is to make your idea dynamic or adaptive for different kind (size) of input frames, and easy to be used for users. Thanks!

tszumski: You are right, user need to prepare such file.
We have project/solution that create regions by neural network and object tracking with SVT-HEVC integrated (we send segmenOv without addtional files)
The whole concept for reading from file is just to prove that this enhancement works, and as a example how to integrate/use this feature

Source/Lib/Codec/EbCodingLoop.c
tszumski: "EB_BOOL useDeltaQp"  parameter is set when separate blocks("LCU") have different value than frame QP. Those changes works similarly to improveSharpness/bitRateReduction. Even if this is just part of seg OV feature, checking segmentOvEnabled is sufficient and more save.

Source/API/EbApi.h
tszumski: Fixed by implementation of 2nd idea ("SegmentOvFile")

Source/API/EbApi.h
tszumski: Fixed by implementation of 2nd idea ("SegmentOvFile")

tszumski:
Hi guys, any update on this? We'd appreciate it if we could merge it to main-stream :)

tianjunwork:
Hi @tszumski , thank you for the contribution of this feature!

PR #404, opened 2019-10-29
Matched the default configuration parameters between README and encoder.

### LINKED ISSUE #403: Some configuration parameters mismatch between README and encoder

Austin-Hu:
Fixes #403 .

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -3282,7 +3282,7 @@ EB_ERRORTYPE EbH265EncInitParameter(
     configPtr->improveSharpness = EB_FALSE;
 
     // Bitstream options
-    configPtr->codeVpsSpsPps = 0;
+    configPtr->codeVpsSpsPps = 1;
     configPtr->codeEosNal    = 0;
     configPtr->switchThreadsToRtPriority = EB_TRUE;
     configPtr->fpsInVps      = EB_TRUE;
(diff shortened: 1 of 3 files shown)


Source/Lib/Codec/EbEncHandle.c
```diff
@@ -3282,7 +3282,7 @@ EB_ERRORTYPE EbH265EncInitParameter(
     configPtr->improveSharpness = EB_FALSE;
 
     // Bitstream options
-    configPtr->codeVpsSpsPps = 0;
+    configPtr->codeVpsSpsPps = 1;
     configPtr->codeEosNal    = 0;
```
tianjunwork: please verify ffmpeg plugin

tianjunwork: App enabled, not the library.

Austin-Hu: Even though codeVpsSpsPps is set to 1 (as default) in encoder, it wouldn't have the side effect to insert redundant VPS/SPS/PPS. Becuase in SVT-HEVC ffmpeg plugin, codeVpsSpsPps will be reset to 0 when AV_CODEC_FLAG_GLOBAL_HEADER is set into the bitmap of AVCodec due to AVFMT_GLOBALHEADER is assigned for most containers (such as mp4 implemented by movenc).

So keep codeVpsSpsPps with 1 as default in encoder.

Austin-Hu: BTW, keep codeVpsSpsPps with 1, in case that EbH265EncStreamHeader() wouldn't be invoked by any application to insert VPS/SPS/PPS.

PR #406, opened 2019-10-31
Fix integer wrap-around in fillerBytes for CBR VBV.

phamd:
Set fillerBytes to 0 if it would wrap-around to max uint32. This will end up sending FILLER_DATA_OVERHEAD(6) bytes as filler still with no ff_bytes (which is valid).


1480c1:
Please run
```bash
git commit --amend -s
git push --force
```
to sign off on your commit

intelmark:
Looks fine..

PR #409, opened 2019-10-31
Fix condition to prevent encoding two AUD in one IDR frame.

phamd:
The first AUD is inserted here:
```c
        if (sequenceControlSetPtr->staticConfig.codeVpsSpsPps && toInsertHeaders) {
            // Reset the bitstream before writing to it
            ResetBitstream(
                pictureControlSetPtr->bitstreamPtr->outputBitstreamPtr);

            if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter) {

                EncodeAUD(
                    pictureControlSetPtr->bitstreamPtr,
                    pictureControlSetPtr->sliceType,
                    pictureControlSetPtr->temporalId);
            }
```
So the second AUD now checks if the first one was not inserted.

### DIFF
--- a/Source/Lib/Codec/EbPacketizationProcess.c
+++ b/Source/Lib/Codec/EbPacketizationProcess.c
@@ -570,8 +570,7 @@ void* PacketizationKernel(void *inputPtr)
         packetizationQp = pictureControlSetPtr->pictureQp;
 
         // Note: If AUD is inserted with headers, avoid insertion again for smaller bitstream.
-        if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter && pictureControlSetPtr->pictureNumber > 0
-                && !toInsertHeaders)
+        if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter && !toInsertHeaders)
         {
             EncodeAUD(
                 pictureControlSetPtr->bitstreamPtr,
--- a/Source/Lib/Codec/EbPacketizationProcess.c
+++ b/Source/Lib/Codec/EbPacketizationProcess.c
@@ -568,8 +568,10 @@ void* PacketizationKernel(void *inputPtr)
 
         // Encode slice header and write it into the bitstream.
         packetizationQp = pictureControlSetPtr->pictureQp;
-     
-        if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter && !(sequenceControlSetPtr->staticConfig.codeVpsSpsPps && toInsertHeaders))
+
+        // Note: If AUD is inserted with headers, avoid insertion again for smaller bitstream.
+        if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter && pictureControlSetPtr->pictureNumber > 0
+                && !toInsertHeaders)
         {
             EncodeAUD(
                 pictureControlSetPtr->bitstreamPtr,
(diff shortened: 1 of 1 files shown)


Source/Lib/Codec/EbPacketizationProcess.c
```diff
@@ -569,7 +569,7 @@ void* PacketizationKernel(void *inputPtr)
         // Encode slice header and write it into the bitstream.
         packetizationQp = pictureControlSetPtr->pictureQp;
      
-        if(sequenceControlSetPtr->staticConfig.accessUnitDelimiter && (pictureControlSetPtr->pictureNumber > 0)) 
+        if (sequenceControlSetPtr->staticConfig.accessUnitDelimiter && !(sequenceControlSetPtr->staticConfig.codeVpsSpsPps && toInsertHeaders))
```
Austin-Hu: Hi @phamd ,

I don't think it's a reasonable logic to encode AUD, as it's just a condition mutually exclusive with the previous one in the same PacketizationKernel() function. Actually EbH265EncStreamHeader() would also encode AUD once it's invoked by application. Please reconsider your change. Thanks!

phamd: With `rateControlMode=1` and `intraRefreshType=0`, the current condition `(pictureControlSetPtr->pictureNumber > 0)` isn't mutually exclusive with `toInsertHeaders`; I'm seeing this state with 2 AUD in the output ES. But it could be the case that there's a different bug in this block:

```c
        toInsertHeaders = EB_FALSE;
        if (pictureControlSetPtr->pictureNumber == 0) {
            toInsertHeaders = EB_TRUE;
        } else if ((pictureControlSetPtr->sliceType == EB_I_PICTURE) &&
                   (sequenceControlSetPtr->intraRefreshType >= IDR_REFRESH)) {
            if (sequenceControlSetPtr->staticConfig.rateControlMode) {
                EB_U32 idrCount = pictureControlSetPtr->pictureNumber /
                                  (sequenceControlSetPtr->intraPeriodLength + 1);
                if ((idrCount % (sequenceControlSetPtr->intraRefreshType + 1)) == 0)
                    toInsertHeaders = EB_TRUE;
            } else if (sequenceControlSetPtr->intraRefreshType == IDR_REFRESH) {
                toInsertHeaders = EB_TRUE;
            }
        }
```

Austin-Hu: Hi @phamd ,

I was saying, your change is just alternative condition compared with Line 260 in this file, but maybe it's not a reasonable logic to trigger AUD encoded. Thanks!

tianjunwork:
Hi @phamd, I didn't reproduce two AUD before/after one IDR frame. Could you let me know the command line triggers the issue you see?

phamd:
Command line to reproduce: 
`SvtHevcEncApp -i /dev/zero -b out.265 -w 1920 -h 1080 -fps 5 -n 20 -ua-delm 1 -rc 1 -irefresh-type 0`
First IDR frame: AUD->VPS->SPS->PPS->IDR_slice
Following IDR frames: AUD->VPS->SPS->PPS->AUD->IDR_slice

tianjunwork:
There is no issue for decoder(HM etc.) to understand bitstream with the extra AUD. And it doesn't violate the spec.
Remove it for the unnecessary extra bits.

With the fix:
fix

Master:
old

PR #422, opened 2019-11-15
Safe string functions refactor

tianjunwork:
Port over from SVT-AV1
Author: John Stebbins <jstebbins@jetheaddev.com>

- Consolidate safe string functions in App
- Duplicate string functions in lib with different names
These functions are not desired in the lib's API, but are used in both
the lib and the App. They can't have the same name since this causes
duplicate symbols in statically link executable.


Source/App/EbAppString.h
```diff
@@ -0,0 +1,172 @@
...
+#ifndef ESNOTFND
+#define ESNOTFND        ( 409 )       /* not found                   */
+#endif
+
+#define RSIZE_MAX_STR      ( 4UL << 10 )      /* 4KB */
+#define RCNEGATE(x)  (x)
+
+#ifndef sldebug_printf
+#define sldebug_printf(...)
+#endif
```
1480c1: `sldebug_printf` gets defined to nothing, so `sldebug_printf("IGNORE CONSTRAINT HANDLER: (%u) %s\n", error, (msg) ? msg : "Null message");` gets converted to ` `

tianjunwork:
Hi Mark @intelmark, EbString.c and .h is copied from AV1, but function names are changed to eb_hevc_* to avoid potential conflict with other svt libraries.

PR #424, opened 2019-11-19
YUV444 10bit compressed

intelmark:
1) Removed configuration check that disallowed YUV444 10 bit compressed
configuration

2) Change the LCU stride for the YUV444 incremental buffer  chroma from
32 to 64


Austin-Hu:
@intelmark , please add the "Signed-off-by" line.

PR #438, opened 2019-12-06
Terminate without core dump running with illegal instruction

tianjunwork:

### DIFF
--- a/Source/App/EbAppMain.c
+++ b/Source/App/EbAppMain.c
@@ -53,10 +53,14 @@ extern APPEXITCONDITIONTYPE ProcessOutputStreamBuffer(
 
 volatile int32_t keepRunning = 1;
 
-void EventHandler(int32_t dummy) {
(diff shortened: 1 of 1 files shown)


tianjunwork:
Tested with compiling on xeon with avx512, then copy binaries over to a avx2 only machine. It works.

intelmark:
Good improvement.  I guess there's no way to print out which instruction sets the app was actually compiled with?

xwu2git:
The original request that the library should not crash is a result from using SVT within a container environment. A container can run on any platforms. Core dump is not acceptable as a result. Thus the app, the ffmpeg plugin or gst plugin should not core dump. Especially for the plugins, a core dump brings down the entire app. Intercepting SIGILL is not a solution as it is already too late (at least for the plugins.) The plugins can be absent due to unqualified platform but should not bring down the entire pipeline. 

Certainly this patch addresses the encoder app issue alone. That's ok but not sufficient to address the original issue.

Issue #357, opened 2019-09-10
SVT-HEVC multi-instance are blocked when calling deinit by multi-thread


Yeongjin-Jeong:
I used the gstreamer svthevcenc plugin, and found that if I create a multi-instance by allocating 1 instance per 1 thead, sometimes it would blocked when calling deinit per thread. So they are not terminated. This occurred when cpu utilization was very high.

Does SVT-HEVC support multi-instance by multi-thread in single process? I have checked that multi-instance works fine by single thread.

The command used is:
```
gst-launch-1.0 videotestsrc num-buffers=1800 ! video/x-raw,width=1920,height=1080,framerate=30/1 ! queue ! tee name=t \
t. ! queue ! svthevcenc speed=9 key-int-max=30 bitrate=6000 ! h265parse ! queue ! fakesink \
t. ! queue ! svthevcenc speed=9 key-int-max=30 bitrate=6000 ! h265parse ! queue ! fakesink \
... (branches for high cpu utilization)
```

tianjunwork:
Hi @Yeongjin-Jeong , just to confirm, what you mean "multi-instance" is "The multi-instance functionality" mentioned in the README or the way you use the plugin?
The multi-instance functionality is an application level feature, one SvtHevcEncApp can launch multiple encoder instances by using -nch N.

Yeongjin-Jeong:
That "multi-instance" means multiple svthevcenc elements. I used gstreamer for creating multiple outputs of the same input in the same process. Each svthevcenc element uses libSvtHevcEnc through its own thread.

Yeongjin-Jeong:
@gokul989, If you want to run on single thread, don't use the queue element before the svthevcenc element.

Such as:
* single process with multi thread
```
gst-launch-1.0 videotestsrc num-buffers=1800 ! video/x-raw,width=1920,height=1080,framerate=30/1 ! queue ! tee name=t \
t. ! queue ! svthevcenc speed=9 key-int-max=30 bitrate=6000 ! h265parse ! queue ! fakesink \
t. ! queue ! svthevcenc speed=9 key-int-max=30 bitrate=6000 ! h265parse ! queue ! fakesink \
... (many branches for high cpu utilization)
```
* single process with single thread
```
gst-launch-1.0 videotestsrc num-buffers=1800 ! video/x-raw,width=1920,height=1080,framerate=30/1 ! queue ! tee name=t \
[...]

PR #446, opened 2019-12-11
Fixed the memory map management issue with multi-instance.

Austin-Hu:
Because currently almost all the objects allocated in heap are
managed by the memoryMap, addressed the issue by:
1. Managing the memory maps with encoding context via linked list;
2. Searching for the corresponding memory map with the current
   encoding context.

Note: thread ID (POSIX pthread_t) can't be used to distinguish the
memory maps, because the SVT APIs may be invoked within different
threads by applications or framework plugins.

NOTE: the most reasonable solution is:
1. Free the allocated objects at the end of each kernel thread;
2. Join each kernel thread, destruct each kernel context and System
   Resource object in EbDeinitEncoder().

Fixes Issue #357 .


Austin-Hu:
Don't need it any more, as PR #467 targets to address this issue as well.

PR #454, opened 2019-12-24
Gstreamer plugin: Update several properties to gstreamer style

Yeongjin-Jeong:
* Update serveral properties to gstreamer style
  - Change property name
  - Separate `intra-refresh` property to explicit action properties
  - Use GEnum instead of constant value
* Fix invalid property id
* Fix invalid maximum value for `speed` property
* Add new properties for vbv factor
* Remove redundant gst git-patch
  - Since the upstreaming of the gstreamer-plugin is done, we no longer need to keep git-patch in this repository.
  - See https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/merge_requests/731


tianjunwork:
Thank you again Yeongjin for the many fixes to the gst plugin.
I am in vacation. Will work on this PR when back to work. But at the first glance, it looks great!

Yeongjin-Jeong:
@tianjunwork, Thanks for your attention! I'll continue to make effort for maintenance. Happy new year :)

PR #460, opened 2020-01-17
Multithreading of Unpack2D

PierreIntel:
Multithreading the Unpack2D function to remove bottleneck for non compressed 10 bits files in resolutions greater than 4k.

### DIFF
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -39,6 +39,7 @@
 #include "EbEncDecProcess.h"
 #include "EbEntropyCodingProcess.h"
 #include "EbPacketizationProcess.h"
+#include "EbUnPackProcess.h"
 
 #include "EbResourceCoordinationResults.h"
 #include "EbPictureAnalysisResults.h"
@@ -117,6 +118,9 @@ EB_U32                         libMallocCount = 0;
 EB_U32                         libThreadCount = 0;
 EB_U32                         libSemaphoreCount = 0;
 EB_U32                         libMutexCount = 0;
+
+EB_U32                          nbLumaThreads = 1;
+EB_U32                          nbChromaThreads = 1;
 
 EB_U8                           numGroups = 0;
 #ifdef _WIN32
@@ -519,6 +523,7 @@ static EB_ERRORTYPE EbEncHandleCtor(
     encHandlePtr->encDecThreadHandleArray                           = (EB_HANDLE*) EB_NULL;
     encHandlePtr->entropyCodingThreadHandleArray                    = (EB_HANDLE*) EB_NULL;
     encHandlePtr->packetizationThreadHandle                         = (EB_HANDLE) EB_NULL;
+    encHandlePtr->unpackThreadHandleArray                           = (EB_HANDLE*) EB_NULL;
 
     // Contexts
     encHandlePtr->resourceCoordinationContextPtr                    = (EB_PTR) EB_NULL;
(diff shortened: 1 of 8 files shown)


PierreIntel:
I duplicate the failed test but I didn't have any errors.
I tried with gcc 5.4.0, 8.3.0, -DCMAKE_BUILD_TYPE, -DBUILD_SHARED_LIBS=OFF and with the command line : 
./SvtHevcEncApp -encMode 9 -i akiyo_cif.y4m -b test.h265 -lp 2 (-asm 0)

1480c1:
Error is that the CPU they used was too slow so the test did not finish in time

Austin-Hu:
Hi @PierreIntel ,

I also reproduced the random hang [issue](https://travis-ci.com/OpenVisualCloud/SVT-HEVC/jobs/276753021) at least once within 20 times, with and without the PR #460 applied, by the command line:
```./SvtHevcEncApp -i akiyo_cif.y4m -b test1.h265 -lp 1
```

So the hang issue isn't due to your PR #460 , which looks good to me, and indeed improves the encoding performance for big resolutions. So please make sure that the playback of encoded bit streams is correct for other formats (YUV422/444) as well, and then we will go through the review and merging process for the PR. Thanks!


Hi @intelmark , @1480c1 , @tianjunwork ,

I checked that the random hang issue exists for long time since v1.3.0, with the akiyo_cif.y4m file or other YUV low resolution inputs (such as 352x288), at the stack:
```
#0  0x00007ffff764b6d6 in futex_abstimed_wait_cancelable (private=0, abstime=0x0, expected=0, futex_word=0x555559f2b3f0) at ../sysdeps/unix/sysv/linux/futex-internal.h:205
#1  do_futex_wait (sem=sem@entry=0x555559f2b3f0, abstime=0x0) at sem_waitcommon.c:111
#2  0x00007ffff764b7c8 in __new_sem_wait_slow (sem=0x555559f2b3f0, abstime=0x0) at sem_waitcommon.c:181
#3  0x00007ffff78b8f59 in EbBlockOnSemaphore () from libSvtHevcEnc.so.1
#4  0x00007ffff78b8c33 in EbGetEmptyObject () from libSvtHevcEnc.so.1
#5  0x00007ffff787c6e7 in EbH265EncSendPicture () from libSvtHevcEnc.so.1
#6  0x000055555555ea25 in ProcessInputBuffer ()
#7  0x0000555555557ea4 in main ()
```

Once you also confirm it, we need to open an issue to track and fix it. Thanks!

PierreIntel:
Hello @Austin-Hu , I verified the output files with an input in 422/444 and the option color-format 2/3, there are no errors and I can playback them correctly.

Source/Lib/Codec/EbEncHandle.c
```diff
@@ -3621,42 +3710,72 @@ static EB_ERRORTYPE CopyFrameBuffer(
...
+            unpackDataPtr->outnBitBuffer       = inputPicturePtr->bufferBitIncCr + chromaBufferOffset  + (EB_U32)(out_chroma/nb_chroma)*i;
+            unpackDataPtr->outnStride          = inputPicturePtr->strideBitIncCr;
+            unpackDataPtr->width               = chromaWidth;
+            unpackDataPtr->height              = chromaHeight/nb_chroma;
+            EbPostFullObject(copyFrameBufferWrapperPtr);
+        }
+        
+        EbObjectWrapper_t   *unpackEndSyncWrapperPtr;
+        for(int i = 0 ; i < nb; i++){
+            EbGetFullObject(context_unpack->unPackOutPutFifoPtr,&unpackEndSyncWrapperPtr);
```
Austin-Hu: Can we wait for the object posted from the unpackTasksConsumerFifoPtrArray queue, to make sure the buffers are completely unpacked (synced)? I mean, maybe we don't need the redundant unpackSync SystemResource and producer/consumer queues? Thanks!

PierreIntel: I think we can't wait for it in CopyFrameBuffer because the unpackTasksConsumerFifo is a one way queue to deliver the datas to unpack function. We need a synchronization between CopyFrameBuffer and the threads so the unpackSyncConsumerFifo is doing the sync.

PR #480, opened 2020-02-24
Optimize MaxNumReorderPics in vps/sps

intelmark:
Changed the algorithm used to set the MaxNumReorderPics
written to the bitstream (both in SPS and VPS structures)

NumReorderPics indicates the maximum number of pictures that can precede
any picture in decoding order and follow it in output order.  The value
needs to be no greater than the DPB size

Added DetermineNumReorderPics to calculate the value based on the
prediction structure in the configuration.  For Low Delay we set
NumReorderPics to 0 and for Random Acces we set NumReorderPics to
a value according to the number of hierarchical levels

### DIFF
--- a/Source/Lib/Codec/EbEntropyCoding.c
+++ b/Source/Lib/Codec/EbEntropyCoding.c
@@ -5344,9 +5344,9 @@ static void CodeProfileTierLevel(
 }
 
 /**************************************************
-* DetermineMax
+* ComputeNumReorderPics
 **************************************************/
-EB_U32 DetermineNumReorderPics(SequenceControlSet_t    *scsPtr)
+EB_U32 ComputeNumReorderPics(SequenceControlSet_t    *scsPtr)
 {
     // NumReorderPics indicates the maximum number of pictures that can precede any picture in decoding 
     // order and follow it in output order.  The value needs to be no greater than the DPB size
@@ -5438,11 +5438,11 @@ static void CodeVPS(
 		    bitstreamPtr,
 			scsPtr->maxDpbSize - 1/*4*/);//(1 << (scsPtr->maxTemporalLayers-syntaxItr)));
 
-        // Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
-        // decoding order and follow it in output order.
-        WriteUvlc(
-            bitstreamPtr,
-            DetermineNumReorderPics(scsPtr));
+		// Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
+		// decoding order and follow it in output order.
+		WriteUvlc(
+			bitstreamPtr,
+			ComputeNumReorderPics(scsPtr));
 
         // Max Latency Increase Plus1
 		WriteUvlc(
@@ -6063,11 +6063,11 @@ static void CodeSPS(
 			bitstreamPtr,
 			scsPtr->maxDpbSize - 1/*4*/);//(1 << (scsPtr->maxTemporalLayers-syntaxItr)));
         
-        // Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
-        // decoding order and follow it in output order.
-        WriteUvlc(
-            bitstreamPtr,
-            DetermineNumReorderPics(scsPtr));
+		// Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
+		// decoding order and follow it in output order.
+		WriteUvlc(
+		bitstreamPtr,
+		ComputeNumReorderPics(scsPtr));
 
 		WriteUvlc(
 			bitstreamPtr,
--- a/Source/Lib/Codec/EbEntropyCoding.c
+++ b/Source/Lib/Codec/EbEntropyCoding.c
@@ -5344,6 +5344,28 @@ static void CodeProfileTierLevel(
 }
 
 /**************************************************
+* DetermineMax
+**************************************************/
+EB_U32 DetermineNumReorderPics(SequenceControlSet_t    *scsPtr)
+{
+    // NumReorderPics indicates the maximum number of pictures that can precede any picture in decoding 
+    // order and follow it in output order.  The value needs to be no greater than the DPB size
+    //
+    // The reorderPicsPerHierarchy values are determined by inspection of the graphs at the top of 
+    // shown at the top of EbPredictionStructure.c and determining the highest observed value for 
+    // NumReorderPics in any one picture within that prediction structure.
+    //                    hierarchical level    1  2  3  4  5   6
+    static EB_U32 reorderPicsPerHierarchy[] = { 0, 1, 2, 4, 6, 14 };
+
+    if (scsPtr->predStructPtr->predType == EB_PRED_LOW_DELAY_P || scsPtr->predStructPtr->predType == EB_PRED_LOW_DELAY_B)
+        // For low delay prediction structures set num reorder pictures to 0
+        return 0;
+    else
+        // For Random Access prediction structures set num reorder pictures according to the number of hierarchical layers
+        return MIN(scsPtr->maxDpbSize - 1, reorderPicsPerHierarchy[scsPtr->maxTemporalLayers]);
+}
+
+/**************************************************
 * CodeVPS
 **************************************************/
 static void CodeVPS(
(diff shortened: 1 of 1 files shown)


Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5343,6 +5343,28 @@ static void CodeProfileTierLevel(
 	return;
 }
 
+/**************************************************
+* DetermineMax
+**************************************************/
+EB_U32 DetermineNumReorderPics(SequenceControlSet_t    *scsPtr)
```
tianjunwork: Recommend: DetermineNumReorderPics ==> ComputeNumReorderPics, which seems the convention in SVT.
Also internal CI stopped. Pls run CI fast with this changes locally(Note, needs install libs), to check if the HM decoder runs fine.

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5412,14 +5434,17 @@ static void CodeVPS(
 	{
 		// Supports Main Profile Level 5 (Max Picture resolution 4Kx2K, up to 6 Hierarchy Levels)
 		// Max Dec Pic Buffering [i]
-		WriteUvlc(
-			bitstreamPtr,
+	    WriteUvlc(
+		    bitstreamPtr,
 			scsPtr->maxDpbSize - 1/*4*/);//(1 << (scsPtr->maxTemporalLayers-syntaxItr)));
 
+		// Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
```
tianjunwork: pls keep comment same as in spec: vps_max_num_reorder_pics to make it easier to search in the code.
Also keep the same style as other comments.

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5412,14 +5434,17 @@ static void CodeVPS(
 	{
 		// Supports Main Profile Level 5 (Max Picture resolution 4Kx2K, up to 6 Hierarchy Levels)
 		// Max Dec Pic Buffering [i]
-		WriteUvlc(
-			bitstreamPtr,
+	    WriteUvlc(
```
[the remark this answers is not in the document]
tianjunwork: pls fix it again.

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -6037,10 +6062,12 @@ static void CodeSPS(
...
 			bitstreamPtr,
 			scsPtr->maxDpbSize - 1/*4*/);//(1 << (scsPtr->maxTemporalLayers-syntaxItr)));
-
+        
+		// Max Num Reorder pics - indicates the maximum number of pictures that can precede any picture in
+		// decoding order and follow it in output order.
 		WriteUvlc(
-			bitstreamPtr,
-			scsPtr->maxDpbSize - 1/*3*/);//(1 << (scsPtr->maxTemporalLayers-syntaxItr)));
+		bitstreamPtr,
```
tianjunwork: pls fix the coding style.

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5343,6 +5343,28 @@ static void CodeProfileTierLevel(
...
+EB_U32 ComputeNumReorderPics(SequenceControlSet_t    *scsPtr)
+{
+    // NumReorderPics indicates the maximum number of pictures that can precede any picture in decoding 
+    // order and follow it in output order.  The value needs to be no greater than the DPB size
+    //
+    // The reorderPicsPerHierarchy values are determined by inspection of the graphs at the top of 
+    // shown at the top of EbPredictionStructure.c and determining the highest observed value for 
+    // NumReorderPics in any one picture within that prediction structure.
+    //                    hierarchical level    1  2  3  4  5   6
+    static EB_U32 reorderPicsPerHierarchy[] = { 0, 1, 2, 4, 6, 14 };
```
tianjunwork: do you need it to be static? always visible? and it doesn't need to be U32.

tianjunwork: I guess your original thought is `const`?

intelmark: maxDpbSize and maxTemporalLayers are both U32 so might be best to keep the same as all three are used in the MIN macro

I can change to const and drop the static since it's inside a function.  Note: I don't think we're consistent on the use of constant throughout the code  (e.g. the prediction structures at the top of EbPredictionStructure are static (not const) ).

tianjunwork: We totally have 4-Level Hierarchies.
And level 4 has 3 NumReorderPics.
EB_U32 reorderPicsPerHierarchy[] = { 0, 1, 2, 3 };

Source/Lib/Codec/EbEntropyCoding.c
```diff
@@ -5343,6 +5343,28 @@ static void CodeProfileTierLevel(
...
+
+    if (scsPtr->predStructPtr->predType == EB_PRED_LOW_DELAY_P || scsPtr->predStructPtr->predType == EB_PRED_LOW_DELAY_B)
+        // For low delay prediction structures set num reorder pictures to 0
+        return 0;
+    else
+        // For Random Access prediction structures set num reorder pictures according to the number of hierarchical layers
+        return MIN(scsPtr->maxDpbSize - 1, reorderPicsPerHierarchy[scsPtr->maxTemporalLayers]);
```
tianjunwork: no need to use MIN() here.
And your thought is for safety reason, it should be `MIN(scsPtr->maxDpbSize, reorderPicsPerHierarchy[scsPtr->maxTemporalLayers]);`, since here it is `vps_max_num_reorder_pics` not `vps_max_num_reorder_pics_minus1`

Source/Lib/Codec/EbEntropyCoding.c
intelmark: In level 4 hierarchical (referring to the graph in EbPredictionStructure.cpp), picture 5 has 4 pictures which precede it decoding order (3,2,4,1)  but follow it in output order.  That is the highest value of NumReorderPics I can see.

intelmark: We expose 4 hierarchical levels in the configuration interface, but the prediction structure code defines 6 levels (i.e. PredictionStructureConfigArray in EbPredictionStructure.cpp which is used to enumerate the number of levels in the PredictionStructureGroup constructor (PredictionStructureGroupCtor).

Source/Lib/Codec/EbEntropyCoding.c
intelmark: The function calculates the value of max_num_reorder_pics (both for sps_max_num_reorder_pics and vps_max_num_reoder_pics.  Per the spec:

sps_max_num_reorder_pics[ i ] indicates the maximum allowed number of pictures with PicOutputFlag equal to 1 that can precede any picture with PicOutputFlag equal to 1 in the CVS in decoding order and follow that picture with PicOutputFlag equal to 1 in output order when HighestTid is equal to i. The value of sps_max_num_reorder_pics[ i ] shall be in the range of 0 to sps_max_dec_pic_buffering_minus1[ i ], inclusive. 

The value we are writing to the bit stream for sps_max_dec_pic_buffer_minus1 is scsPtr->maxDpbSize - 1, hence the check for in the MIN statement to make sure the value doesn't exceed  scsPtr->maxDpbSize-1 .

Source/Lib/Codec/EbEntropyCoding.c
tianjunwork: Assume the number you are talking about is `display order` in line 175, pic 6 has pic 6,8 decode before it but display after it.

Source/Lib/Codec/EbEntropyCoding.c
tianjunwork: I see, thx. But I think MIN is not needed.

Source/Lib/Codec/EbEntropyCoding.c
intelmark: On line 175 picture 1 in display order, needs pictures 2, 4, 6, and 8 decoded first, and these follow it in output order.  I think 4 makes sense for the MaxNumReorderPics value for this hierarchy.

Source/Lib/Codec/EbEntropyCoding.c
intelmark: CI Fast has been run locally without problems.  Also encoded files have been viewed in VQ Analyzer to confirm they can be decoded and SPS/VPS value for MaxNumReorderPics is appropriate.

Source/Lib/Codec/EbEntropyCoding.c
tianjunwork: picture 1 doesn't need 6. pls check again. vicue is a better view.

PR #482, opened 2020-02-27
Fixed the SegFault issue for 64x64 fractional search in ME.

### LINKED ISSUE #481: SVT HEVC encoder crashes with the particular HEVC standard test sequence SlideShow
latest code crashes at line 660 of EbPictureOperators_C.c with the below command. asm code has the same issue.

-i SlideShow_1280x720.yuv -w 1280 -h 720 -asm 0 -encMode 0 -intra-period 100 -scd 0 -rc 0 -q 22 -n 100 -b SlideShow_1280x720_SVTMODE0_22_SVT.265

Austin-Hu:
The issue happens when 64x64 fractional search is enabled with
encMode 0~2. Occationally the calculated xSearchIndex in
PU_HalfPelRefinement() could be less than 0. So getting the Y
pixels of reference buffer (integerBufferPtr of ME context) can
be done by forward shift, rather than indexing. Because
integerBufferPtr locates at some middle region region of the PA
reference buffer's Y component, and it's "legal" to access its
preceding memory region.

Fixes #481 .

### DIFF
--- a/Source/Lib/Codec/EbMotionEstimation.c
+++ b/Source/Lib/Codec/EbMotionEstimation.c
@@ -799,7 +799,7 @@ static void PU_HalfPelRefinement(
         *pBestSsd = (EB_U32) SpatialFullDistortionKernel_funcPtrArray[!!(ASM_TYPES & PREAVX2_MASK)][Log2f(puWidth) - 2](
             &(contextPtr->lcuSrcPtr[puLcuBufferIndex]),
             contextPtr->lcuSrcStride,
-            refBuffer + (ySearchIndex * (EB_S32)refStride + xSearchIndex),
+            &(refBuffer[ySearchIndex * refStride + xSearchIndex]),
             refStride,
             puHeight,
             puWidth);
--- a/Source/Lib/Codec/EbMotionEstimation.c
+++ b/Source/Lib/Codec/EbMotionEstimation.c
@@ -799,7 +799,7 @@ static void PU_HalfPelRefinement(
         *pBestSsd = (EB_U32) SpatialFullDistortionKernel_funcPtrArray[!!(ASM_TYPES & PREAVX2_MASK)][Log2f(puWidth) - 2](
             &(contextPtr->lcuSrcPtr[puLcuBufferIndex]),
             contextPtr->lcuSrcStride,
-            &(refBuffer[ySearchIndex * refStride + xSearchIndex]),
+            refBuffer + (ySearchIndex * (EB_S32)refStride + xSearchIndex),
             refStride,
             puHeight,
             puWidth);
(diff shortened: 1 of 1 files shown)


Austin-Hu:
I'm using the latest [commit](https://github.com/OpenVisualCloud/SVT-HEVC/commit/ddbc81c4b0497b6a47d393ebc7d77a19ccd51323) of master branch plus PR #482 , with the same command line you reported, but I didn't see crashing here with several times running:
```
./SvtHevcEncApp -i SlideShow_1280x720.yuv -w 1280 -h 720 -asm 0 -encMode 0 -intra-period 100 -scd 0 -rc 0 -q 22 -n 100 -b 0.265
```

BTW, as I mentioned in the commit comments, (ySearchIndex * (EB_S32)refStride + xSearchIndex) could be negative number, but refBuffer points to some middle range of the refPicPtr->bufferY buffer, so the data ahead of it can be accessed.

Austin-Hu:
Temporarily worked around this issue by resetting best MV for each LCU, and will continue tracing its root cause. BTW, the bitstreams are exactly the same after encoding other YUV inputs.

Source/Lib/Codec/EbMotionEstimation.c
```diff
@@ -799,7 +799,7 @@ static void PU_HalfPelRefinement(
         *pBestSsd = (EB_U32) SpatialFullDistortionKernel_funcPtrArray[!!(ASM_TYPES & PREAVX2_MASK)][Log2f(puWidth) - 2](
             &(contextPtr->lcuSrcPtr[puLcuBufferIndex]),
             contextPtr->lcuSrcStride,
-            &(refBuffer[ySearchIndex * refStride + xSearchIndex]),
+            refBuffer + (ySearchIndex * (EB_S32)refStride + xSearchIndex),
```
intelmark: @Austin-Hu  - did you find this change made any difference?

Austin-Hu: It just works around invalid addressing for some LCUs, and actually it's unecessary with the 2nd commit applied. So it will be removed once the solution is finalized.

Austin-Hu:
> Presuming encoding speed was the same with the fix as well as the bitstream being identical?

Yes. But right now the change hasn't been confirmed as a fix, and it's not the right time to compare encoding performance.

tianjunwork:
My fix is `#define  MAX_SAD_VALUE 64*64*255` ==> `#define  MAX_SAD_VALUE 64*64*255+1`, similar to Mark's fix. With this change, FullPelSearch_LCU guarantees to update the mv. 64 is the max lcu size, 255 is the max SAD of a pixel.

PDX-Runner:
has Hassene confirmed that this is the right fix or it is still a temporal work around? if we know how much padding area is allocated for the reference frame buffer, can we simply clamp the ME search area to be inside the reference frame buffer + padding area?

tianjunwork:
Hi @PDX-Runner , currently there are two proposal to fix: 1. init the mv before doing full pixel search(otherwise, mv is same as the previous one), 2. adjust SAD value to make sure mv is updated for sure during full pixel search.
Both of them are not confirmed by Hassene. We will look into it today still to find the root cause.

PR #485, opened 2020-03-04
Remove setting of FPS

intelmark:
When processing y4m header information, the FPS information
should be ignored as it controls the playback of the file, but should
not impact the encoding.  This commit removes the setting of
the configuration framerate when processing the y4m header.


tianjunwork:
Can we just comment out the lines instead of deleting?

intelmark:
What would be the reason for allowing them to still be visible?

tianjunwork:
> What would be the reason for allowing them to still be visible?


It is not a big deal. But visually show that we ignore the FPS setting in the header.

PR #487, opened 2020-03-05
Clean up useless function

tianjunwork:


tianjunwork:
Originally had the change in #467, removed from it based on code review.

PR #493, opened 2020-03-13
appveyor: update pacman packages along with installing them

1480c1:
Fixes issue in appveyor with packages not being installed because of gcc


tianjunwork:
Thank you @1480c1 for the quick fix!

Issue #501, opened 2020-03-20
Inconsistent encodings between asm 0 and 1 with AVX512 enabled 


intelmark:
Found an issue with inconsistent encodings (different sized bitstreams) when using a 10-bit y4m file as input and when the number of frames requested is larger than input.  This appears to be specific to y4m files.

The file causing problems is NetflixTango_4096x2160_60p_i010_10bit.y4m and has 294 frames.  
When the encoding command line requests more than 294 frames, the resulting bitstreams are different.

(e.g.)
SvtHevcEncApp/SvtHevcEncApp -encMode 10 -i /testfiles/benchmarking/NetflixTango_4096x2160_60p_i010_10bit.y4m -bit-depth 10 -n 290 -nb -1 -asm 0 -b NetflixTango_4096x2160_60p_i010_10bit_ENCODEMODE_10_SVT-HEVC_asm0.hevc

SvtHevcEncApp/SvtHevcEncApp -encMode 10 -i /testfiles/benchmarking/NetflixTango_4096x2160_60p_i010_10bit.y4m -bit-depth 10 -n 290 -nb -1 -asm 1 -b NetflixTango_4096x2160_60p_i010_10bit_ENCODEMODE_10_SVT-HEVC_asm1.hevc
Resulting hevc files are the same

SvtHevcEncApp/SvtHevcEncApp -encMode 10 -i /testfiles/benchmarking/NetflixTango_4096x2160_60p_i010_10bit.y4m -bit-depth 10 -n 300 -nb -1 -asm 0 -b NetflixTango_4096x2160_60p_i010_10bit_ENCODEMODE_10_SVT-HEVC_asm0.hevc

SvtHevcEncApp/SvtHevcEncApp -encMode 10 -i /testfiles/benchmarking/NetflixTango_4096x2160_60p_i010_10bit.y4m -bit-depth 10 -n 300 -nb -1 -asm 1 -b NetflixTango_4096x2160_60p_i010_10bit_ENCODEMODE_10_SVT-HEVC_asm1.hevc
Resulting hevc files are different

Also duplicated with Netflix_Aerial_1920x1080_60fps_8bit_420_60f_10bit.y4m

PR #503, opened 2020-03-23
y4m file rewind logic

intelmark:
Core problem is that the existing assert and return of
EB_ErrorNone when the y4m frame delimiter is read
are happening when the file stream reader is at the end of
the file.  This can happen when the -n value is larger than the
actual number of frames in the file.

To fix this problem with y4m files, an additional check is
added to when the y4m frame delimiter is being read.  Rather
than asserting after the feof check, the y4m file is rewound, the
header is skipped over, and the frame delimiter is attempted to be
read again.

Note: One alternative solution could be to first figure out how many
bytes are in the file (using fseek and ftell), and then use ftell
to see if the file reader is at the end of the file.

This logic would replace the feof call at the end of ReadInputFrames,
since that call is not correctly being used. feof uses an indicator that
is set by a previous operation on the stream that attempted to read
at or past the end-of-file. By itself calling feof does not return any
useful information.

### DIFF
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -314,8 +314,18 @@ int32_t read_y4m_frame_delimiter(EbConfig_t *cfg) {
     fresult = fgets((char *)bufferY4Mheader, sizeof(bufferY4Mheader), cfg->inputFile);
 
     if (fresult == NULL) {
-        assert(feof(cfg->inputFile));
-        return EB_ErrorNone;
+        // Could be at end of file
+        // Rewind file and skip over header
+        fseek(cfg->inputFile, 0, SEEK_SET);
+        char buffer[YFM_HEADER_MAX];
+        fresult = fgets(buffer, sizeof(buffer), cfg->inputFile);
+
+        // try reading delimiter again
+        fresult = fgets((char *)bufferY4Mheader, sizeof(bufferY4Mheader), cfg->inputFile);
+        if (fresult == NULL) {
+            assert(feof(cfg->inputFile));
+            return EB_ErrorNone;
+        }
     }
     if (!validateAlphanumeric(bufferY4Mheader)){
         return EB_ErrorBadParameter;
--- a/Source/App/EbAppInputy4m.c
+++ b/Source/App/EbAppInputy4m.c
@@ -5,6 +5,7 @@
 
 #include "EbAppInputy4m.h"
 #define YFM_HEADER_MAX 80
+#define YFM_FRAME_DELIMITER_MAX 10
 
 #define CHROMA_MAX 4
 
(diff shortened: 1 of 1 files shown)


tianjunwork:
closes #501

Source/App/EbAppInputy4m.c
```diff
@@ -314,8 +314,18 @@ int32_t read_y4m_frame_delimiter(EbConfig_t *cfg) {
...
-        return EB_ErrorNone;
+        // Could be at end of file
+        // Rewind file and skip over header
+        fseek(cfg->inputFile, 0, SEEK_SET);
+        char buffer[YFM_HEADER_MAX];
+        fresult = fgets(buffer, sizeof(buffer), cfg->inputFile);
+
+        // try reading delimiter again
+        fresult = fgets((char *)bufferY4Mheader, sizeof(bufferY4Mheader), cfg->inputFile);
+        if (fresult == NULL) {
```
tianjunwork: btw, bufferY4Mheader naming is misleading, which is actually the frame header. `Each frame begins with the 5 bytes 'FRAME' followed by zero or more parameters each preceded by 0x20, ending with 0x0A. This is then followed by the raw bytes from each plane.`

Source/App/EbAppInputy4m.c
```diff
@@ -314,8 +314,18 @@ int32_t read_y4m_frame_delimiter(EbConfig_t *cfg) {
     fresult = fgets((char *)bufferY4Mheader, sizeof(bufferY4Mheader), cfg->inputFile);
 
     if (fresult == NULL) {
-        assert(feof(cfg->inputFile));
-        return EB_ErrorNone;
+        // Could be at end of file
+        // Rewind file and skip over header
+        fseek(cfg->inputFile, 0, SEEK_SET);
+        char buffer[YFM_HEADER_MAX];
+        fresult = fgets(buffer, sizeof(buffer), cfg->inputFile);
```
tianjunwork: Can it be just fseek by size of YFM_HEADER_MAX? You don't actually use `buffer`.

Source/App/EbAppInputy4m.c
intelmark: Agreed.. I can easily rename the variable

Source/App/EbAppInputy4m.c
intelmark: The header is read with fgets:
char * fgets ( char * str, int num, FILE * stream );

fgets reads characters from stream and stores them as a C string into str until (num-1) characters have been read or either a newline or the end-of-file is reached, whichever happens first.

The buffer is allocated to a maximum size (YFM_HEADER_MAX = 80), but fewer than that number of bytes are actually read.  The number of bytes read varies from file to file because the width, height, and other values are stored as ascii strings separated by tokens.

tianjunwork: I see, thx.

PR #510, opened 2020-03-31
Perf improvment with threads affinity

PierreIntel:
Signed-off-by: PierreIntel <pierre.vallee@intel.com>


tianjunwork:
Please add detailed description of your PR in the **commit message**.
And speed/memory/quality impact in this **PR page**.

PierreIntel:
It didn't change the memory or the quality, only the performance.
It give a gain of ~15% with a Intel Xeon 8280 with "-ss -111111110" in the command line.
This load all steps on socket 1 and encDec, unpack and Main threads on socket 0.

tianjunwork:
@PierreIntel `-ss -111111110` I don't get the value for `-ss`. It should be within [-2 - 1] from the code `SVT_LOG("SVT [Error]: Instance %u: Invalid TargetSocket [-2 - 1]`.
`-ss` originally means the ID of socket. `-1` is the default non-configed value. What does `-2` mean? Could you be specific about the API change? Do you have a better design of the interface?

deeptiag1:
ss flag was originally meant to do socket affinity, In this PR, the flag has been modified do thread affinity for the process. The ss flag now consists of 9(numbers/chars). Each char specifies the socket which will  be used by each process(1 for 1 process, except enc-dec) 
Lp flag was to specify the number of cores. In this PR Lp has been modified to select the number of threads for encdec on a socket.
For eg., if you have 24 enc dec threads(as per SVT code).  To pin 24 on socket 1 the parameter is 124, 1-> indicates socket 1, 24->indicates 24 threads for enc dec.
Let us know if this works. If not, please guide us the right way to code it.

tianjunwork:
1. EB_CREATE_THREAD_SOCKET only support Linux. How about windows support? Even if we don't need to have this feature on windows, we can't break windows running.
2. Although the PR modification of -ss and -lp is backward compatible, it is not a good idea to change the purpose of the interface completely. And it makes testing complicated.
How about adding a new interface? Leave -ss and -lp unchanged.

PR #517, opened 2020-04-07
pkg-config: use CMAKE_INSTALL_FULL for ${prefix}

heijligen:


1480c1:
I would not recommend using the full version because of the standard convention being to use the prefix variable so the installer can just sed the prefix line to change where they installed it

1480c1:
I guess maybe one use of this would be if you have your libdir and includedir not under prefix

heijligen:
I'm packaging for nixos and there @CMAKE_INSTALL_INCLUDEDIR@ is an absolute path. So I get 
```
prefix=/nix/store/cakcpnxl0w9w2gjg406s3f5avyngyyih-svt-hevc-1.4.3
includedir=/nix/store/72jx1mdxddald3m2w6863542mk7lqghd-svt-hevc-1.4.3-dev//nix/store/72jx1mdxddald3m2w6863542mk7lqghd-svt-hevc-1.4.3-dev/include/svt-hevc
libdir=${prefix}//nix/store/cakcpnxl0w9w2gjg406s3f5avyngyyih-svt-hevc-1.4.3/lib
```
when using @CMAKE_INSTALL_FULL_INCLUDEDIR@ as the svt-av1 lib does everything went fine.

1480c1:
I see, so it is installing to a dir not under prefix

heijligen:
correct

1480c1:
> correct. Nix does not use the linux filesystem hierarchy standard

I see, then can you mention that reason in the commit message? Something like `to accommodate non-standard install locations` or similar, since right now, it's not easy to understand why this change was made without looking at this pull request

PR #475, opened 2020-02-14
Fixed the encoding random hang issue with few threads.

Austin-Hu:
This is an old issue since v1.2.0. Overall it's due to the incorrect
life cycle management of the PPCS, PA RefPic, PCS objects in different
kernels (through ResourceCoordination to EncDec, and PictureManger &
RateControl with feedbacks from EncDec and Packetization separately).

When the random hang happens, the former process is:
1. PPCS A which is referenced by other picture(s) B (& C), has been
   encoded and packetized completely. RateControl releases it after
   getting the feedback back from Packetization;
2. The PPCS A object is enqueued into the head of (the empty queue of)
   PPCS FIFO (for higher cache hit);
3. ResourceCoordination receives a new picture D, dequeues the PPCS
   object (aligned with A) from head of PPCS FIFO, and makes picture D
   go through the latter encoding stages;
4. In PictureDecision, the contents (including isUsedAsReferenceFlag)
   of the PPCS object are updated according to parameters of picture D;
5. Meanwhile in EncDec, PPCS A is still on-going with its latter stuff
   (such as SAO, padding, etc.) after encoding and feedbacking to
   EntropyCoding. But as the isUsedAsReferenceFlag of PPCS object
   belonged to A was changed, A couldn't be fed back to PictureManager.
6. In the inputPictureQueue of PictureManger, picture B (& C) are
   permanently waiting for A to be available;
7. Encoding for latter pictures is blocked (objects from FIFO queues
   exhaust) and hangs.

So the solution is:
(Note: we need to manage the life cycle of PPCS, PA RefPic and PCS
RefPic objects separately, because they're operated in
pictureDecisionPaReferenceQueue of PictureDecision and
referencePictureQueue of PictureManager, besides of other kernels.)

1. In ResourceCoordination, make sure the live count of PPCS and its PA
   RefPic objects are 2, because they will be asynchronously released
   in both EncDec and RateControl kernels;
2. In PictureDecision, increase the dependentPicturesCount of the PA
   RefPic object, when a PPCS is indeed referenced by other picture(s).
   And don't release PPCS and its PA RefPic objects at the end when
   trying to clean up the entries of the pictureDecisionPaReferenceQueue;
3. In InitialRateControl, don't release PPCS and its PA RefPic objects,
   as they'd be used in the following kernels;
4. In PictureManager, increase the refCount of the PCS RefPic object,
   when the PCS indeed references other picture(s);
5. In EncDec, release PPCS and its PA RefPic objects which the encoded
[...]

PR #525, opened 2020-04-16
Replaced memcpy with referencing input frame buffers.

Austin-Hu:
By replacing memcpy in EbH265EncSendPicture, with direct assigning the
pointers of Input FIFO object with the ones of the EB_H265_ENC_INPUT
object YUV buffers, which should be safe (loked) by application.

The buffers belonging to an Input FIFO object don't need to be padded,
because they're just used as source sample data in PictureAnalysis,
MotionEstimation and EncDec (sharpness enabled) kernels.

Note: width and height of input frames are required to be aligned with 8
samples by de-blocking filter. So the solution applies for almost all typical
resolutions.

ffmpeg & GStreamer SVT-HEVC plugin: references the input AVFrame objects, and
manages the input referenced frames with linked list.

### DIFF
--- a/Source/API/EbApi.h
+++ b/Source/API/EbApi.h
@@ -10,7 +10,6 @@
 #include <string.h>
 #include <stdint.h>
 #include <EbApiVersion.h>
-#include <sys/queue.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -101,8 +100,6 @@ extern "C" {
         uint8_t* naluBase64Encode;
 
         SegmentOverride_t *segmentOvPtr;
-
-        LIST_ENTRY(EB_BUFFERHEADERTYPE) list;
     } EB_BUFFERHEADERTYPE;
 
     typedef struct EB_COMPONENTTYPE
--- a/Source/API/EbApi.h
+++ b/Source/API/EbApi.h
@@ -10,6 +10,7 @@
 #include <string.h>
 #include <stdint.h>
 #include <EbApiVersion.h>
+#include <sys/queue.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -100,6 +101,8 @@ extern "C" {
         uint8_t* naluBase64Encode;
 
         SegmentOverride_t *segmentOvPtr;
+
+        LIST_ENTRY(EB_BUFFERHEADERTYPE) list;
     } EB_BUFFERHEADERTYPE;
 
     typedef struct EB_COMPONENTTYPE
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -1912,18 +1912,42 @@ void EbHevcSetParamBasedOnInput(
         sequenceControlSetPtr->videoUsabilityInfoPtr->frameFieldInfoPresentFlag = EB_TRUE;
     }
 
-    sequenceControlSetPtr->maxInputPadRight = 0;
-    sequenceControlSetPtr->maxInputPadBottom = 0;
-
-    sequenceControlSetPtr->lumaWidth    = sequenceControlSetPtr->maxInputLumaWidth;
-    sequenceControlSetPtr->lumaHeight   = sequenceControlSetPtr->maxInputLumaHeight;
-
     // Temporarily not to reference the input frames directly when sharpness improvement
     // is enabled. Because some frames have been completedly encoded and returned to
     // application to be unreferenced, but they may be still used to generate the
     // reconstructed picture in EncDec.
     sequenceControlSetPtr->refInputFrame =
         !sequenceControlSetPtr->staticConfig.improveSharpness ? EB_TRUE : EB_FALSE;
+
+    // And not to reference the input frames directly once either width or height
+    // is not aligned to 8 samples.
+    sequenceControlSetPtr->refInputFrame =
+        ((sequenceControlSetPtr->maxInputLumaWidth % MIN_CU_SIZE) || (sequenceControlSetPtr->maxInputLumaHeight % MIN_CU_SIZE)) ?
+        EB_FALSE : sequenceControlSetPtr->refInputFrame;
+
+    // Update picture width, and picture height
+    if (sequenceControlSetPtr->maxInputLumaWidth % MIN_CU_SIZE) {
+
+        sequenceControlSetPtr->maxInputPadRight = MIN_CU_SIZE - (sequenceControlSetPtr->maxInputLumaWidth % MIN_CU_SIZE);
+        sequenceControlSetPtr->maxInputLumaWidth = sequenceControlSetPtr->maxInputLumaWidth + sequenceControlSetPtr->maxInputPadRight;
+        //sequenceControlSetPtr->maxInputChromaWidth = sequenceControlSetPtr->maxInputLumaWidth >> 1; //TODO: change here
+    }
+    else {
+
+        sequenceControlSetPtr->maxInputPadRight = 0;
+    }
+    if (sequenceControlSetPtr->maxInputLumaHeight % MIN_CU_SIZE) {
+
+        sequenceControlSetPtr->maxInputPadBottom = MIN_CU_SIZE - (sequenceControlSetPtr->maxInputLumaHeight % MIN_CU_SIZE);
+        sequenceControlSetPtr->maxInputLumaHeight = sequenceControlSetPtr->maxInputLumaHeight + sequenceControlSetPtr->maxInputPadBottom;
+        //sequenceControlSetPtr->maxInputChromaHeight = sequenceControlSetPtr->maxInputLumaHeight >> 1; //ditto
+    }
+    else {
+        sequenceControlSetPtr->maxInputPadBottom = 0;
+    }
+
+    sequenceControlSetPtr->lumaWidth    = sequenceControlSetPtr->maxInputLumaWidth;
+    sequenceControlSetPtr->lumaHeight   = sequenceControlSetPtr->maxInputLumaHeight;
 
     // Configure the padding
     if (sequenceControlSetPtr->refInputFrame) {
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -1918,7 +1918,12 @@ void EbHevcSetParamBasedOnInput(
     sequenceControlSetPtr->lumaWidth    = sequenceControlSetPtr->maxInputLumaWidth;
     sequenceControlSetPtr->lumaHeight   = sequenceControlSetPtr->maxInputLumaHeight;
 
-    sequenceControlSetPtr->refInputFrame = EB_TRUE;
+    // Temporarily not to reference the input frames directly when sharpness improvement
+    // is enabled. Because some frames have been completedly encoded and returned to
+    // application to be unreferenced, but they may be still used to generate the
+    // reconstructed picture in EncDec.
+    sequenceControlSetPtr->refInputFrame =
+        !sequenceControlSetPtr->staticConfig.improveSharpness ? EB_TRUE : EB_FALSE;
 
     // Configure the padding
     if (sequenceControlSetPtr->refInputFrame) {
--- a/Source/Lib/Codec/EbEncHandle.c
+++ b/Source/Lib/Codec/EbEncHandle.c
@@ -3886,14 +3886,6 @@ EB_ERRORTYPE AllocateFrameBuffer(
         (EB_PTR)&inputPictureBufferDescInitData);
     inputBuffer->pBuffer = (uint8_t*)buf;
 
-    if (is16bit && config->compressedTenBitFormat == 1) {
-        const EB_COLOR_FORMAT colorFormat = (EB_COLOR_FORMAT)sequenceControlSetPtr->chromaFormatIdc;
-        //pack 4 2bit pixels into 1Byte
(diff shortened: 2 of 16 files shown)


Source/API/EbApi.h
```diff
@@ -10,6 +10,7 @@
 #include <string.h>
 #include <stdint.h>
 #include <EbApiVersion.h>
+#include <sys/queue.h>
```
1480c1, line 13: Does this header exist on windows and mac?

Austin-Hu: No, they should use another library of doubly-linked list. The change is a prototype tried in Linux, and need to be also applied to Windows/Mac if it's accepted. Thanks!

Austin-Hu:
Encoding performance and memory consumption comparison between master [commit](https://github.com/OpenVisualCloud/SVT-HEVC/commit/b2530415115efeb492923cc3d0604cf2b070a9b9) and with the PR applied, when encoding 8K 8-bit P420 input:

| Command Line | [Master](https://github.com/OpenVisualCloud/SVT-HEVC/commit/b2530415115efeb492923cc3d0604cf2b070a9b9) | | PR #475 | |
| -- | -- | -- | -- | -- |
| | FPS | Memory (GB) | FPS | Memory (GB) |
| ./**SvtHevcEncApp** -i 8k2D_musician_7680x3840_8bit_30Hz_P420.yuv -w 7680 -h 3840 -b 0.265 -n 80000 | 40.35 | 44.91 | **40.90** | **38.05** |
| ./**ffmpeg** -stream_loop -1 -video_size 7680x3840 -r 60 -i 8k2D_musician_7680x3840_8bit_30Hz_P420.yuv -c:v libsvt_hevc -y 0.265 | 25 | 44.91 | **30** | **38.05** |

Got the encoding performance data, after encoding ~50000 frames. And there is little improvement by encoding with SvtHevcEncApp which needs more investigation.

tianjunwork:
Thank you very much Austin for prototyping this feature!

Source/API/EbApi.h
1480c1: Seems that there is something similar for windows, but not exactly https://docs.microsoft.com/en-us/windows/win32/api/ntdef/ns-ntdef-list_entry

Austin-Hu: Thanks! I checked it as well, and it seems that it's Windows kernel mode library, but not sure whether we could apply it in user mode. And I don't know the common one used in Mac...

Here is the clear [implementation](https://github.com/torvalds/linux/blob/master/include/linux/list.h) of doubly-linked list in Linux kernel.

BTW, I looked for such library in libavutil of ffmpeg, but didn't find it.

Source/Lib/Codec/EbEncHandle.c
```diff
@@ -1936,15 +1936,25 @@ void EbHevcSetParamBasedOnInput(
...
-
     sequenceControlSetPtr->lumaWidth    = sequenceControlSetPtr->maxInputLumaWidth;
     sequenceControlSetPtr->lumaHeight   = sequenceControlSetPtr->maxInputLumaHeight;
 
+    if (!(sequenceControlSetPtr->lumaWidth % MAX_LCU_SIZE) && !(sequenceControlSetPtr->lumaHeight % MAX_LCU_SIZE))
+        sequenceControlSetPtr->lcuAligned = EB_TRUE;
+
+    // Configure the padding
+    if (sequenceControlSetPtr->lcuAligned) {
+        sequenceControlSetPtr->leftPadding  = 0;
```
hassount, line 1947: even if it's LCU aligned, we would need the extra padding for subpel. This might not show up in fast presets, rather in the better quality presets

Source/Lib/Codec/EbEncHandle.c
```diff
@@ -3365,27 +3375,37 @@ static EB_ERRORTYPE CopyFrameBuffer(
 
 
         // Y
-        for (inputRowIndex = 0; inputRowIndex < lumaHeight; inputRowIndex++) {
-
-            EB_MEMCPY((inputPicturePtr->bufferY + lumaBufferOffset + lumaStride * inputRowIndex),
-                (inputPtr->luma + sourceLumaStride * inputRowIndex),
-                lumaWidth);
+        if (sequenceControlSetPtr->lcuAligned)
+            inputPicturePtr->bufferY = inputPtr->luma;
```
hassount, line 3379: this assumes that the layout of the input buffer is what the library is expecting, such might be the case with the svt sample app but not other applications as decoder buffers might use a different padding structure.

Austin-Hu: Yes, because the ffmpeg libsvt_hevc plugin (based on pure software encoder) doesn't implemente the callbacks of HWContextType, like what other hardware acceleration encoders do, ffmpeg doesn't know the buffer requirement (such as stride alignment) of SVT-HEVC encoder. And need to check whether GStreamer has the similar behavior as well.

Source/API/EbApi.h
```diff
@@ -100,6 +101,8 @@ extern "C" {
         uint8_t* naluBase64Encode;
 
         SegmentOverride_t *segmentOvPtr;
+
+        LIST_ENTRY(EB_BUFFERHEADERTYPE) list;
```
tianjunwork, line 105: Since there is no utility you can use on Windows, and the linked list is only used in app layer. Let's write our own linked list in app layer to manage the buffer.

SGShen:
I did a test with the Sample App. The PR improves FPS and encoding time, somehow, it causes latency almost doubled. Has anyone noticed the similar issue? Thanks.


  | encMode | FPS | Encoding time | Exec Time | Avg Latency | Max Latency
-- | -- | -- | -- | -- | -- | --
no PR | 10 | 39.29 | 25450 | 29144 | 396 | 562
w/ PR | 10 | 47.71 | 20962 | 24680 | 796 | 1210
no PR | 9 | 29.11 | 34357 | 38027 | 1064 | 1550
w/ PR | 9 | 30.25 | 33059 | 36755 | 2405 | 3168


The command line I used is: 
$numactl --cpubind=0 --membind=0 ./SvtHevcEncApp -i ~/tests/videos/7680x3840_420_5.yuv -w 7680 -h 3840 -fps 30 -scd 0 -pred-struct 0 -irefresh-type 0 -rc 1 -tbr 70000000 -temporal-id 0  -lad 4 -encMode 10 -fpsinvps 0 -tile_col_cnt 6 -tile_row_cnt 12 -tile_slice_mode 1 -intra-period 4 -n 1000

Austin-Hu:
> I did a test with the Sample App. The PR improves FPS and encoding time, somehow, it causes latency almost doubled. Has anyone noticed the similar issue? Thanks.
> 
>  	encMode	FPS	Encoding time	Exec Time	Avg Latency	Max Latency
> no PR	10	39.29	25450	29144	396	562
> w/ PR	10	47.71	20962	24680	796	1210
> no PR	9	29.11	34357	38027	1064	1550
> w/ PR	9	30.25	33059	36755	2405	3168
> The command line I used is:
> $numactl --cpubind=0 --membind=0 ./SvtHevcEncApp -i ~/tests/videos/7680x3840_420_5.yuv -w 7680 -h 3840 -fps 30 -scd 0 -pred-struct 0 -irefresh-type 0 -rc 1 -tbr 70000000 -temporal-id 0 -lad 4 -encMode 10 -fpsinvps 0 -tile_col_cnt 6 -tile_row_cnt 12 -tile_slice_mode 1 -intra-period 4 -n 1000

Hi @SGShen ,

We’d better encode for more (> 10000) frames after the statistics are stable enough.

The “Avg Latency” is calculated in SvtHevcEncApp by totalLatency / frameCount, where totalLatency is accumulated from the “encoding time” of each frame. But, the problem is that the “encoding time” is measured since the 1st encoding stage (Resource Coordination) to the last one (Packetization), but doesn’t contain the original memcpy time taken in EbH265EncSendPicture which is the FIRST point to encode each frame. So please get the latency in App level since you call EbH265EncSendPicture, and to EbH265GetPacket where you get the corresponding encoded frame (with PTS). Thanks!

SGShen:
@Austin-Hu Understood the memcpy was not counted in the "Avg latency". But the results showed there is a negative impact of this PR on "Avg latency". 

Like your mentioned, "Avg. latency" = "Total Latency"/ "frame counts", and "Total Latency" does not include memcpy which is removed from this PR. So it seems unlikely to get a worse "Avg. Latency" with this PR. Either the calculation formula is incorrectly executed, or there is a hidden impact to the encoding.

Austin-Hu:
With the 4th [commit](https://github.com/OpenVisualCloud/SVT-HEVC/pull/525/commits/29e6136a90af0c4b6cc4e79a4cd8a5ecbe927327) applied, here is the updated encoding performance and memory consumption comparison between master and with the PR applied, when encoding 8-bit P420 inputs with ffmpeg (encMode 7 & tune 1 as default) on Intel Xeon 8280:

| Resolution | Command Line | FPS | | Memory (GB) | |
| -- | -- | -- | -- | -- | -- |
| | | Master | PR #525 | Master | PR #525 |
| 480p | ./ffmpeg -stream_loop -1 -video_size 720x480 -r 60 -i   8k2D_musician_720x480_8bit_30Hz_P420.yuv -c:v libsvt_hevc -y 0.265 | 641 | **676** | 1.72 | **1.64** |
| 1080p | ./ffmpeg -stream_loop -1 -video_size 1920x1080 -r 60 -i Fallout4_1920x1080_8bit_60Hz_P420.yuv -c:v libsvt_hevc -y 0.265 | 222 | **223** | 4.4 | **4.01** |
| 4K | ./ffmpeg -stream_loop -1 -video_size 3840x2160 -r 60 -i 8k2D_musician_3840x2160_8bit_30Hz_P420.yuv -c:v libsvt_hevc -y 0.265 | 138 | **152** | 13.5 | **11.48** |
| 8K | ./ffmpeg -stream_loop -1 -video_size 7680x4320 -r 60 -i 8k2D_musician_7680x4320_8bit_30Hz_P420.yuv -c:v libsvt_hevc -y 0.265 | 23 | **27** | 49.39 | **41.7** |

ffmpeg_plugin/0001-lavc-svt_hevc-add-libsvt-hevc-encoder-wrapper.patch
```diff
@@ -510,6 +568,49 @@ index 0000000000..fce27a06ea
...
++        while (input_frame_entry) {
++            if (input_frame_entry->ref_frame->pts == pkt->pts) {
++                // Unference the AVFrame oject which has been encoded completely.
++                av_frame_unref(input_frame_entry->ref_frame);
++                free(input_frame_entry);
++
++                if (QueryDepthSList(&svt_enc->tmp_input_frame_list) &&
++                        last_entry && first_entry) {
++                    // first_entry has been pushed to the last entry of tmp_input_frame_list.
++                    InterlockedPushListSListEx(&svt_enc->input_frame_list, last_entry,
```
1480c1, line 587: Seems that using `InterlockedPushListSListEx` requires at least Windows 8, and defining `NTDDI_VERSION` to at least `0x06020000`. There seems to be [`InterlockedPushListSList`](https://docs.microsoft.com/en-us/windows/win32/sync/interlockedpushlistslist) which instead requires the windows version to be higher than windows xp (Which in turn gets defined to `InterlockedPushListSListEx` if the windows version is greater than 8)
```c
libavcodec/libsvt_hevc.c: In function 'eb_encode_frame':
libavcodec/libsvt_hevc.c:501:21: error: implicit declaration of function 'InterlockedPushListSListEx'; did you mean 'RtlInterlockedPushListSListEx'? [-Werror=implicit-function-declaration]
  501 |                     InterlockedPushListSListEx(&svt_enc->input_frame_list, last_entry,
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                     RtlInterlockedPushListSListEx
```

Austin-Hu: Thanks, @1480c1! As using InterlockedPushListSList still caused the compiling issue, I've to drop it with another way instead.

Austin-Hu:
If sharpness is enabled, this PR depends on another one #531.

Austin-Hu:
Even though memcpy for input frame takes “much” (~15 ms for 8K input) time for each frame, it takes much longer time (no big difference with different “-pred-struct” values) when each frame goes through the whole encoding pipeline, as what “Average Latency” shows after running SvtHevcEncApp:

| Command Line | Time Consumed By Each Frame (ms) | | |
| -- | -- | -- | -- |
| | **CopyFrameBuffer** | **Encoding Pipeline** | |
| | | LAD 17 | LAD 1 |
| ./SvtHevcEncApp -i   Fallout4_1920x1080_8bit_60Hz_P420.yuv -w 1920 -h 1080 -n 10000 -b 0.265 | ~1 | ~700 | ~680 |
| ./SvtHevcEncApp -i rally_8192x4096_30.yuv -w 8192 -h 4096 -n 10000 -b 0.265 | ~15 | ~3000 | ~2400 |

<br/>

So even though CopyFrameBuffer takes \~0 ms after applying this PR, it couldn’t help to improve the encoding performance. And this PR would only be helpful, in **low latency mode** where encoding a frame takes much less time (for example, 100~200 ms).

intelmark:
Given the limited use case of this PR, the team thought best to close it for now.  In most cases the CopyFrameBuffer percentage to overall encoding time is quite low, hence the total latency is not reduced much.

PR #527, opened 2020-04-19
App/EbAppContext.c: Remove unused AllocateOutputBuffers

1480c1:
Since nothing is using the `streamBufferPool`, it would be better to remove it for now

Matching SVT-AV1 PR: https://github.com/OpenVisualCloud/SVT-AV1/pull/1209


tianjunwork:
Thank you Chris porting it over to SVT-HEVC/VP9.

PR #530, opened 2020-04-27
Update the Y sample buffer indexing for position A in Motion Estimation.

Austin-Hu:
NxM SAD is calculated between the current LCU buffer and its corresponding
reference one in reconstructed picture for O/A/B/C/D positions.

Note: the same change should be also applied in hme_mv_center_check()
      of SVT-AV1.


Austin-Hu:
Need to check whether the missed updating of searchRegionIndex for position A is because of the algorithm, or a bug.

tianjunwork:
Hi @Austin-Hu It is a bug. Thank you for the PR.

PR #531, opened 2020-04-27
Update refDenSrcPicture source as reconstructed buffer.

Austin-Hu:
Because after handling all the segments of a picture in EncDec kernel,
and after generating the reconstructed buffer in EncodePass() when
it's referenced, the input buffer is no longer needed and can be
returnd to application. The source data for either SAO or
refDenSrcPicture preparation (when sharpness is enabled) can be from
reconstructed buffer.


Austin-Hu:
Close the incorrect fix, because refDenSrcPicture is used to store the original content of input buffer, as the reference picture to encode latter pictures.

PR #534, opened 2020-05-14
Adjust PAREF_OUT

intelmark:
Adjusted the PAREF_OUT PR 455
+ Moved the padding of the input picture prior to the copying of the
picture to the picture analysis reference picture
+ Removed the PAREF_OUT #define

Testing has shown this will fix the SVT defield test cases which
were failing because the field dimensions were not a multiple of 8.


tianjunwork:
Thank you Austin!

Issue #545, opened 2020-06-22
High bit rate hevc still has some mosaics in some frames


LeoKingLong:
1. Encode command：
SvtHevcEncApp -i 2_hd.yuv -w 7680 -h 3840 -fps 30 -intra-period 4 -irefresh-type 0 -vid-info 1 -rc 1 -tbr 62914560 -umv 0 -tile_slice_mode 1 -tile_row_cnt 12 -tile_col_cnt 12 -pred-struct 0 -b 2_hd.hevc

2.Problems :
 Some of the pictures have some mosaics（especially on the face），how to fix this problem？

Thanks!

tianjunwork:
Hi @LeoKingLong , what is the actual bitrate of the bitstream? You need to encode large enough number of frames to get a stable output bitrate. You can use `-n number_of_frames` to repeatedly encode a clip.
If you still see block artifact on the face, could you attach the image?

LeoKingLong:
Hi @tianjunwork 
The original video（mp4 format） information is here：
![image](https://user-images.githubusercontent.com/26246461/85346700-ce23e680-b528-11ea-9685-cc698d55a091.png)

I used ffmpeg command to make a yuv from the original MP4：
ffmpeg -y -i  1.mp4 -vf scale=7680:3840,setsar=1:1  1.yuv

Then I encoded the 1.yuv with SVT.

The image is here:
http://leokinglong.github.io/output7.png
http://leokinglong.github.io/output9.png

tianjunwork:
Hi @LeoKingLong , what is the output bitrate from SVT-HEVC? You can find it in the log print after encoding.
The output is very blocky, not only on face. Could you try with higher bitrate? 
The bitrate is cut half from original AVC bitrate. Actually, HEVC offers from 25% to 50% better data compression at the same level of video quality compared to AVC. Most of the time it is not exactly 50%.

LeoKingLong:
Hi @tianjunwork 
The output bitrate from SVT-HEVC is 48725.12kbps
![image](https://user-images.githubusercontent.com/26246461/85381461-92147400-b570-11ea-8563-2b814baf59ef.png)

Then I make a new hevc :
![image](https://user-images.githubusercontent.com/26246461/85391701-1a991180-b57d-11ea-9de3-c21d2ee16f35.png)

It looks better , thanks

LeoKingLong:
Hi @tianjunwork 
[...]

PR #555, opened 2020-07-03
Add more explanation for -segment-ov parameter

tszumski:
# Description
Add more explanation for -segment-ov parameter
# Issue
#545 

# Author(s)
@tszumski

# Performance impact

- [ ] quality
- [ ] memory
- [ ] speed
- [ ] 8 bit
- [ ] 10 bit
- [x] N/A

# Merge method
- [x] Allow the maintainer to squash and merge when PR is ready to create a 1-commit to the master branch. The maintainer will be able to fix typos / combine commit messages to create a more readable 1-commit message or use whatever is stated in the 'Description' section
- [ ] I will clean up my commits and the maintainer shall use 'rebase and merge' to the master branch

### DIFF
--- a/Config/SVTSegmentOvFile.txt
+++ b/Config/SVTSegmentOvFile.txt
@@ -8,9 +8,8 @@
 # 2. Usage:
 # For each frame in the video, the user can specify regions with better or worse quality.
 # One input frame have to be preceded by the number of regions.
-# Each region consists of 6 numbers (X1, Y1, X2, Y2, data, mode) i.e (X1,Y1)is the top left and (X2,Y2) is the bottom right corner.
+# Each region consists of 6 numbers (X1, Y1, X2, Y2, data, mode) i.e (X1,Y1) is the top left and (X2,Y2) is the bottom right corner. Regions are allowed to overlap in the same frame.
 # Each region is translated into a subset of LCU(largest coding unit) for which the user can override the default qp.
-# Regions can overlap each other, for LCU that this happen valid mode is set by the further region.
 # The mode parameter has to be in range [1..5]. This allows the user to specify what to override by the [data]
 # Mode 1   Set the qp value for each region. [data] is clipped to [0:51] .
 # Mode 2   Change the qp value (i.e increase or decrease by [data]). [data] is clipped to [-25:25] ,
--- a/Config/SVTSegmentOvFile.txt
+++ b/Config/SVTSegmentOvFile.txt
@@ -2,14 +2,47 @@
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #====================== Segment Override File ======================
-# One input frame have to be preceded by the number of regions
-# Each region consist of 6 numbers (X1,Y1,X2,Y2, data, mode) i.e (X1,Y1)top left and (X2,Y2)bottom right corner
-# Mode parameter have to be in range [1..5]. This allow user to specify what to override by the [data]
-#    1.   Set qp value for each region, [data] is clipped to [0:51]
-#    2.   Change qp value i.e increase or decrease by the [data] clipped to [-25:25]
-#    3.   Use this mode to modify not only quality but also blocks splitting, behaviour depends on parameter 1 and 2
-#    4.   Use this mode to modify blocks splitting i.e increase or decrease by the [data] clipped to [-25:25], negative value increase block division
-#    5.   Update first stage of quantization in range [-7;7] positive value reduce bitrate, negative increase
+# 1. Purpose/use case of this feature
+#    This feature is for very specific use-cases like surveillance systems, where dynamic objects such as pedestrians or vehicles are detected and tracked (encoded with higher quality) while the static background is encoded with worse quality.
+#    By encoding only the important/key areas with good quality, the resulting bitstream is smaller and still has high usefulness.
+# 2. Usage:
+# For each frame in the video, the user can specify regions with better or worse quality.
+# One input frame have to be preceded by the number of regions.
+# Each region consists of 6 numbers (X1, Y1, X2, Y2, data, mode) i.e (X1,Y1)is the top left and (X2,Y2) is the bottom right corner.
+# Each region is translated into a subset of LCU(largest coding unit) for which the user can override the default qp.
+# Regions can overlap each other, for LCU that this happen valid mode is set by the further region.
+# The mode parameter has to be in range [1..5]. This allows the user to specify what to override by the [data]
+# Mode 1   Set the qp value for each region. [data] is clipped to [0:51] .
+# Mode 2   Change the qp value (i.e increase or decrease by [data]). [data] is clipped to [-25:25] ,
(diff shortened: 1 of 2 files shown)


Config/SVTSegmentOvFile.txt
```diff
@@ -2,14 +2,43 @@
...
+# 1. Purpose/use case of this feature
+#    This feature is rather for very specific use-case like surveillance systems, where objects like pedestrians or vehicles are detected and tracked (encoded with higher quality) while background is encoded with worse quality.
+#    Thanks to encoding only key areas with good quality bitstream is smaller and still have high value
+# 2. Usage:
+# For each frame in video user can specify regions with better or worse quality.
 # One input frame have to be preceded by the number of regions
 # Each region consist of 6 numbers (X1,Y1,X2,Y2, data, mode) i.e (X1,Y1)top left and (X2,Y2)bottom right corner
+# Region is translated into subset of LCU(largest coding unit) for which user override default qp
 # Mode parameter have to be in range [1..5]. This allow user to specify what to override by the [data]
-#    1.   Set qp value for each region, [data] is clipped to [0:51]
```
intelmark, line 8: # Mode 1   Set the qp value for each region. [data] is clipped to [0:51] .
# Mode 2   Change the qp value (i.e increase or decrease by [data]). [data] is clipped to [-25:25] ,
# Mode 3   Modify not only the quality but also blocks splitting(LCU partition). The behaviour depends on parameter 1 and 2 and the 5th parameter is unused.
# Mode 4   Only modify the block's splitting(LCU partition) (i.e increase or decrease by[data]). [data] is clipped to [-25:25].  Negative value increase block division.
# Mode 5   Update first stage of quantization in range [-7;7] positive value reduce bitrate, negative increase  bitrate.
#
# There are several ways to use mode 3.  You can set:
#    0 0 500 500 35 1 // Set new qp value
#    0 0 500 500 35 3 // Also modify the block splitting.  The 5th value is skipped
#    or
#    0 0 500 500 -5 2 // Increase the block quality by 5
#    0 0 500 500 -5 3 // Also modify block splitting.  The 5th value is skipped
#
# 3. Example:
#    If you want to increase the quality by 10 of an object of size 100x100 which is moving from left to right on the screen, you can use:
#    1    // one region for frame 0
#    0 0 100 100 -10 2
#    1    // one region for frame 1
#    10 0 110 100 -10 2
#    1    // one region for frame 2
#    20 0 120 100 -10 2
#    1    // one region for frame 3
#    30 0 130 100 -10 2
#    1    // one region for frame 4
#    40 0 140 100 -10 2
#    1    // one region for frame 5
#    50 0 150 100 -10 2
#    1    // one region for frame 6
#    60 0 160 100 -10 2**

tszumski:
Hi @intelmark, Thanks for review and comments. I updated description with your suggestions.

Config/SVTSegmentOvFile.txt
```diff
@@ -2,14 +2,44 @@
...
+#    30 0 130 100 -10 2
+#    1    // one region for frame 4
+#    40 0 140 100 -10 2
+#    1    // one region for frame 5
+#    50 0 150 100 -10 2
+#    1    // one region for frame 6
+#    60 0 160 100 -10 2
 3
 0 0 500 500 50 1
 500 500 1000 0 -20 4
```
tianjunwork, line 45: I don't quite understand the example here.
```
3
0 0 500 500 50 1
500 500 1000 0 -20 4
500 500 1000 1000 20 4
2
0 0 500 500 20 1
0 0 300 200 20 1
```
Do it mean
3 regions for frame 0, which is `[0,0] - [500, 500]`, `[500,500] - [1000, 0] is this valid region?`, `[500,500] - [1000,1000]`
2 regions for frame 1, which is `[0,0] - [500, 500]`,  `[0,0] - [300,200]`. why they are overlapped?

What is the rule for the rest of the frames?

tszumski: Hi @tianjunwork 
> 3 regions for frame 0, which is [0,0] - [500, 500], [500,500] - [1000, 0] is this valid region?, [500,500] - [1000,1000]
Correct, 3 regions, all of them are valid
> 2 regions for frame 1, which is [0,0] - [500, 500], [0,0] - [300,200]. why they are overlapped?
They are overlapped to show that this kind of action is allowed

tianjunwork: `(X1,Y1)is the top left and (X2,Y2) is the bottom right corner` [500,500] - [1000, 0] is this valid region?
Could you add `#` line in this example too to explain that region can be overlapped?
Rule for each frames needs to be specified?  In this example, only frame 0 and 1 have rule?  What is the rule for the rest of the frames?

tszumski: > (X1,Y1)is the top left and (X2,Y2) is the bottom right corner [500,500] - [1000, 0] is this valid region?

This is general description, the most important thing is that points in space have to be diagonal to create rectangle

> Rule for each frames needs to be specified? In this example, only frame 0 and 1 have rule? What is the rule for the rest of the frames?

If the number of frames > number of "rules" the application will loop to the beginning of the file

Config/SVTSegmentOvFile.txt
```diff
@@ -2,14 +2,47 @@
...
-#    3.   Use this mode to modify not only quality but also blocks splitting, behaviour depends on parameter 1 and 2
-#    4.   Use this mode to modify blocks splitting i.e increase or decrease by the [data] clipped to [-25:25], negative value increase block division
-#    5.   Update first stage of quantization in range [-7;7] positive value reduce bitrate, negative increase
+# 1. Purpose/use case of this feature
+#    This feature is for very specific use-cases like surveillance systems, where dynamic objects such as pedestrians or vehicles are detected and tracked (encoded with higher quality) while the static background is encoded with worse quality.
+#    By encoding only the important/key areas with good quality, the resulting bitstream is smaller and still has high usefulness.
+# 2. Usage:
+# For each frame in the video, the user can specify regions with better or worse quality.
+# One input frame have to be preceded by the number of regions.
+# Each region consists of 6 numbers (X1, Y1, X2, Y2, data, mode) i.e (X1,Y1)is the top left and (X2,Y2) is the bottom right corner.
```
intelmark, line 11: Each region consists of 6 numbers (X1, Y1, X2, Y2, data, mode) i.e (X1,Y1) is the top left and (X2,Y2) is the bottom right corner.  Regions are allowed to overlap in the same frame.

tianjunwork:
Hi @tszumski , you can push incrementally, so that we can review what is changed. If there is no need to use force push, it should be avoid.

intelmark:
Current SVT-HEVC user guide needs some updating:
Path to segment override file, diffrence between largest and smallest value should be less or equal 25. Note: using this feature will disable ImproveSharpness and BitRateReduction parameters

maybe

_Path to segment override file which will allow for sharpness improvement and bit rate reduction on a per segment basis.  Refer to config/SVTSegmentOvFile.txt for details._

PR #639, opened 2023-04-27
ffmpeg_plugin: rebase documentation against ffmpeg n4.4

jbeich:
- #587 added new lines without updating hunk offsets
- #591 bumped minimum ffmpeg version to `n4.4`
- `n4.4` has https://github.com/ffmpeg/ffmpeg/commit/6accb7718aa4


jbeich:
`n4.4` has https://github.com/ffmpeg/ffmpeg/commit/6accb7718aa4, so supporting `n4.2` or `n4.3` like ffmpeg_plugin/README says is impossible without creating yet another patch.

